Issue #10, opened 2024-11-16
Revert changes to demical places for products and satisfactory entries


Maelstromeous:
As part of the fix for #3, the amounts had to be converted to a float rather than a whole number. Currently items are displaying as 2.00, ideally it should be locale based and displaying as "2" and then decimals should it be required.

Maelstromeous:
Fixed! Thanks to @samsmithnz !

PR #30, opened 2024-11-17
Format numbers with extra decimal places to the max of 2 decimal places, and reduce numbers that don't need the accuracy.

samsmithnz:
This is an effort to format all of the numbers with the same formatting rules. Strip off zeros to 2 decimal places max.

Hopefully it's not controversial - I'm not great with vue, and had to hack a little to get there - I think the result is solid, but no hard feelings if this is not accepted. This partly addresses #10. 

- If the number is 2.0 or 2.00, show `2`
- If the number is 2.5 or 2.50, show `2.5`
- If the number is 2.25 or 2.250, show `2.25`
- If the number is 2.333, show `2.33`

<HR>

**Copilot generated description:**
This pull request focuses on improving the consistency and readability of numerical values displayed in various Vue components by introducing a new utility function, `formatNumber`, and applying it across multiple files. The most important changes include the addition of the `formatNumber` function, and its integration into several components to format numerical values.

### DIFF
--- a/web/src/components/graph/FactoryNode.vue
+++ b/web/src/components/graph/FactoryNode.vue
@@ -25,7 +25,7 @@
               />
               <v-chip class="sf-chip">
                 <game-asset :subject="input.outputPart ?? 'unknown'" type="item" />
-                <span class="ml-2">{{ getPartDisplayName(input.outputPart) }}: {{ input.amount }}/min</span>
+                <span class="ml-2">{{ getPartDisplayName(input.outputPart) }}: {{ formatNumber(input.amount) }}/min</span>
               </v-chip>
             </div>
           </v-col>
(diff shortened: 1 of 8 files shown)


Maelstromeous:
@samsmithnz Not sure this is fully working? Looking at the preview release I'm not seeing it applied? https://satisfactory-factories-67zmlw2dj-maelstromeous-projects.vercel.app/

Screenshot 2024-11-17 at 11 30 55

Maelstromeous:
Think you just might have missed it in the products component for the ingredients, should be all you need

samsmithnz:
Thanks for the feedback. I think it's resolved now, I did a search for ` /min` and `/min` - previously I'd just fixed all the ` /min`. 

Side note - I did this all in Codespaces with Copilot. Really helped me to understand the Vue parts I haven't used before!

Maelstromeous:
Thanks for the correction! Also to answer your todo in the comment, the reason why totalPower is a string is because it's a float, I had some... fun... with that for the total factory power math.

Maelstromeous:
Looks great! Merged, thank you for the contribution!

Issue #35, opened 2024-11-17
Power data errors


Maelstromeous:
![Screenshot 2024-11-17 at 10 25 24](https://github.com/user-attachments/assets/cced19db-53fd-495b-9bad-e98eed00749c)

samsmithnz:
Investigating... 

I believe it's a rounding error in the parsing from `game_data.json`, Cast Screws aren't showing power either, and have a base power requirement of 0.2. Also noticing Steel pipes doesn't show a power if the quantity is low (e.g. 15). 
```
            "building": {
                "name": "constructormk1",
                "power": "0.2"
            },
```

Maelstromeous:
So there is currently a bug with power for when <0.99 of a building is used, thats logged under #5 , however this is a slightly different issue where the recipe itself doesn't appear to be pulling in any power metrics.

But, in the game data there is a power defined:

```json
{
            "id": "Silica",
            "displayName": "Silica",
            "ingredients": [
                {
                    "part": "RawQuartz",
                    "amount": 3,
                    "perMin": 22.5
                }
            ],
            "products": [
                {
                    "part": "Silica",
                    "amount": 5,
                    "perMin": 37.5,
                    "isByProduct": false
                }
            ],
            "building": {
                "name": "constructormk1",
                "power": "0.8"
            },
            "isAlternate": false,
            "isFicsmas": false
        },
```

So I think the site is somehow using the wrong power values, or there's another bug with how power is calculated.

samsmithnz:
In `FactoryInterface.ts`, the TotalPower is a string - shouldn't that be a number? https://copilot-workspace.githubnext.com/Maelstromeous/satisfactory-factories?shareId=32bea2f3-2dd1-42a6-92c3-476a87a56fd4

samsmithnz:
Creating a PR to test. 🚀

Maelstromeous:
Yeah it would be nice if that whole power thing can be cleaned up, I'm not overly happy with it but it "works", but clearly there's some form of bugs with it going on.

samsmithnz:
Looking at it - I think I see the problem. Question: How to I run the unit tests?

Maelstromeous:
I believe that's in the readme if not I need to add it but it's `pnpm test` under /web, there are no tests for other components at this time

samsmithnz:
Thanks - very helpful. I can see a power issue there now and can fix it. Note that on Main one of the tests currently fails.

Maelstromeous:
[...]

PR #40, opened 2024-11-17
Fix power requirements round2

samsmithnz:


web/src/utils/factory-management/buildings.ts
```diff
@@ -21,13 +21,13 @@ export const calculateBuildingRequirements = (factory: Factory, gameData: DataIn
...
-      const buildingCount = (product.amount / productInRecipe.perMin).toFixed(3)
+      const buildingCount = Number((product.amount / productInRecipe.perMin).toFixed(3))
 
       product.buildingRequirements = {
         name: buildingData.name,
         amount: buildingCount,
-        powerPerBuilding: Number(buildingData.power),
-        totalPower: (parseInt(buildingData.power) * parseInt(buildingCount)).toFixed(2),
+        powerPerBuilding: buildingData.power,
+        totalPower:  Number((buildingData.power * buildingCount).toFixed(3)),
```
Maelstromeous, line 30: So is the intention to store totalPower as a string / float? This is the primary reason why I had them as strings in the first place as I couldn't readily use Number type with `toFixed`.

Maelstromeous: Ah never mind, it's wrapped in Number, now why didn't I think of that 😂

samsmithnz:
Aside from one test that fails constantly for me, with the error `Factory with ID 1234 not found.` (How do I resolve this? ), this is ready. 

As discussed in the [issue](#35), it doesn't fix the power issue, but it does improve the code around it.

Maelstromeous:
> Aside from one test that fails constantly for me, with the error Factory with ID 1234 not found. (How do I resolve this? ), this is ready.

The test doesn't actually fail, the code emits a console.error, and it shows up in the test output. It should show 8 passed suites unless you've added some more.

web/src/utils/factory-management/buildings.ts
```diff
@@ -21,13 +21,13 @@ export const calculateBuildingRequirements = (factory: Factory, gameData: DataIn
...
-      const buildingCount = (product.amount / productInRecipe.perMin).toFixed(3)
+      const buildingCount = product.amount / productInRecipe.perMin
 
       product.buildingRequirements = {
         name: buildingData.name,
         amount: buildingCount,
-        powerPerBuilding: Number(buildingData.power),
-        totalPower: (parseInt(buildingData.power) * parseInt(buildingCount)).toFixed(2),
+        powerPerBuilding: buildingData.power,
+        totalPower:  buildingData.power * buildingCount,
```
Maelstromeous, line 30: Nitpick but extra space?

web/src/utils/factory-management/buildings.ts
```diff
@@ -45,18 +45,18 @@ export const calculateBuildingsAndPower = (factory: Factory) => {
...
+        totalPower: 0,
       }
     }
 
     const facBuilding = factory.buildingRequirements[building.name]
 
-    facBuilding.amount = String(parseFloat(facBuilding.amount) + parseFloat(building.amount))
-    facBuilding.totalPower = String(parseFloat(facBuilding.totalPower) + parseFloat(building.totalPower))
+    facBuilding.amount = facBuilding.amount + building.amount
+    facBuilding.totalPower = factory.totalPower + building.totalPower
```
Maelstromeous, line 57: Looks like the behaviour changed here, so it used to use the facBuilding, now it's using the factory total?

web/src/utils/factory-management/products.ts
```diff
@@ -65,11 +65,14 @@ export const calculateProducts = (factory: Factory, gameData: DataInterface) =>
       }
 
       // === Now we need to calculate the parts required per min to make the product ===
-      const productIngredientRatio = product.amount / recipe.products[0].perMin
+      const productIngredientRatio = product.amount / recipe.products[0].perMin // This formular should have no rounding - the decimal points are important here for the floating point calculations
```
Maelstromeous, line 68: Yep valid point, now you've added the formatNumber stuff you did before this can probably be removed. Feel free to remove the comments from this

Maelstromeous (changes requested):

web/src/utils/factory-management/buildings.ts
samsmithnz: Do you know how to nitpick the space right in here? You can do it with the add a suggestion button, and then it's easy to fix right in the PR. 
```suggestion
        totalPower: buildingData.power * buildingCount,
```

web/src/utils/factory-management/products.ts
samsmithnz: Yeh - keep the precision in the code, and format the number for the user in the HTML.

web/src/utils/factory-management/buildings.ts
samsmithnz: I didn't mean to change the behavior - this is a bug and we need to squash it. 🐞 

```suggestion
    facBuilding.totalPower = facBuilding.totalPower + building.totalPower
```

web/src/utils/factory-management/buildings.ts
Maelstromeous: Ahh yes I keep forgetting that's a thing 😄

web/src/utils/factory-management/buildings.ts
samsmithnz: Does that fix it @Maelstromeous?

samsmithnz: Yes - I didn't intend to create the bug, but it's still a bug, and regressions annoy me. :)

Maelstromeous:
Screenshot 2024-11-18 at 12 18 28

Power looks right, buildings though are broken, will create a seperate issue for that fix.

Maelstromeous (approved):

samsmithnz:
What exactly is wrong with the building numbers? Can you give an example? Next after this is definitely more testing! :)

Maelstromeous:
Screenshot bottom right, I should have highlighted it better to be fair. It's showing 10.66666666667x

samsmithnz:
It looks like the Assembler image is missing too? How do I fix that?

samsmithnz:
And now the assembler image is back. Weird. It wasn't in the HTML, but after the fix to the number, it's back! 

image

Maelstromeous:
#49 is likely the culprit, it exists outside of this PR so don't worry about that.

Maelstromeous:
Merged! Thanks for your hard work @samsmithnz! 💪

Issue #31, opened 2024-11-17
Add clone/instansiate button for factories


Lemiort:
I playing almost only with blueprints, so it would be easier for me to place another factory if needed. 
So there are 2 use-cases:

1. Copy existing factory just to increase production
2. Copy and modify to play with numbers

450Volts:
I appreciate the ability to copy a factory, but is there no way of collapsing multiple IDENTICAL factories? I too play exclusively with blueprints, and having 10 crew(100) factories taking 10 factory slots is difficult to manage. Is it possible to just have a [+] [-] button for each factory entry to simply add additional factory multiples to a single factory entry?

Maelstromeous:
@450Volts we've discussed this issue internally, suffice to say at the moment we don't see the value in adding a "multiplier" of a factory, as you could just quite simply update the quantity of the product and invisibily divide it by however many factory BPs you are using.

However, this did bring into question the creation of Factory Folders to place factories within to make the management of the plan easier. That we do see value in, and hopefully is a decent halfway compromise to what you're asking for.

PR #45, opened 2024-11-18
feat: copy factory #31

jeroenpol:
Add the ability to copy a factory.

Fixes https://github.com/Maelstromeous/satisfactory-factories/issues/31


web/src/components/planner/PlannerFactory.vue
```diff
@@ -111,13 +119,13 @@
                   />
                 </span>
                 <span>
-                  <b>{{ getPartDisplayName(part.id) }}</b>: {{ formatNumber(part.amount) }}/min
```
Maelstromeous, line 114: Not sure why these `formatNumber`s were reverted?

Maelstromeous:
Thank you for your PR @jeroenpol! I've left an item for you to review.

web/src/components/planner/PlannerFactory.vue
jeroenpol: Ah, that was a mistake during forking. Fixed!

Maelstromeous:
Also in testing when a factory is copied, the links to other factories don't appear to be updated.

Reproduction:
1. Open the preview site
2. Ask for demo plan
3. Observe that "Copper Basics" has 160 cable exports, which is being imported by Computers.
4. Clone Computers.
5. Observe how Copper Basics is now **not** red, as the links between the two have not been updated correctly.

Once you update a value one of the factories does it then update.

This is because the factories that it dependant upon have not had their data recalculated. You'll need to call `updateFactories` within the `Planner.vue` in order to make this work upon clone.

Maelstromeous (changes requested):
See above

web/src/components/planner/PlannerFactory.vue
```diff
@@ -166,10 +174,10 @@
   import { Factory } from '@/interfaces/planner/FactoryInterface'
   import { DataInterface } from '@/interfaces/DataInterface'
   import { getPartDisplayName } from '@/utils/helpers'
-  import { formatNumber } from '@/utils/numberFormatter'
```
samsmithnz, line 169: If this strips out this import, it will break the page, as the `formatNumber` function requires it. This needs to stay.

jeroenpol: Alright, doing the fix quickly during work wasn't the best idea :P I'll have a look after/before work tomorrow

Maelstromeous: @jeroenpol To assist future collaboration, PRs will now actually build the project's components so these issues are more apparent at dev time.

Maelstromeous:
@jeroenpol could you try not to use force pushes please, we do a squash merge at the end and it really borks with the history of various tools and GitHub in general when the commits they link to don't exist :(

samsmithnz (approved):

Maelstromeous (approved):

jeroenpol:
> @jeroenpol could you try not to use force pushes please, we do a squash merge at the end to keep it nice and tidy, and it really borks with the history of various tools and GitHub in general when the commits they link to don't exist :(

Ah sure! I work in a rebase + amend environment, I'll try to remember for this repo

PR #70, opened 2024-11-18
Add debug buttons to factory cards

### LINKED ISSUE #50: Add debug mode
Need the ability to add a debug mode for easy development and collaboration.

Ideally, simply supplying <siteurl>?debug=true in the URL should do the trick. This should then show buttons on the bottom of each factory to debug it's data in a graphical format without having to go to the local storage, pull out the JSON into your IDE, format it blah blah upon every change.

Ideally if the URL is not the website URL it should show this by default, removing the need to add a query string.

nils-keller-dev:
Closes #50

### DIFF
--- a/web/src/components/planner/PlannerFactoryProducts.vue
+++ b/web/src/components/planner/PlannerFactoryProducts.vue
@@ -152,13 +152,42 @@
     </div>
     <div class="mt-2">
       <v-btn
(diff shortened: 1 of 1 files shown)


Maelstromeous (changes requested):
Just needs to check for the hostname and I think we're good!

Issue #69, opened 2024-11-18
Implement rate limiting on backend API routes


Maelstromeous:
Tracking issue for:
- [ ] https://github.com/Maelstromeous/satisfactory-factories/security/code-scanning/7

PR #74, opened 2024-11-18
Create sharing backend and web service

### LINKED ISSUE #71: Create sharable links
Thinking we could create sharable links for people to create mockups and then share them with other people.

The URL should contain a /share/X endpoint, the X would be a base64 encoded string of their factory data.

In the /share/ endpoint (page) we create a component that is responsible for ingesting the data, overwriting the user's localStorage `factories` data (prompting them if they're sure) and then returning them to the planner.

Or, we create an instance of the planner and merely just load the data into it, keeping the link intact.

Heavily related to #33.

Maelstromeous:
Closes #71 #69 (nice)


samsmithnz:
I'm curious about why you are creating a web service - instead of just letting the web app load and save?

Maelstromeous:
We need to save the factory data somewhere to be shared, which needs the backend service for the database.

Maelstromeous:
No more comments for you sonarqube...

backend/backend.ts
```diff
@@ -155,13 +154,11 @@
 });
 
 // Save Data Endpoint
-app.post('/save', authenticate, async (req: AuthenticatedRequest & TypedRequestBody<{ data: any }>, res: Express.Response) => {
+app.post('/save', authenticate, apiRateLimit, async (req: AuthenticatedRequest & TypedRequestBody<{ data: any }>, res: Express.Response) => {
```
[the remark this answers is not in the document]
Maelstromeous: Well that got it wrong didn't it!

samsmithnz:
> We need to save the factory data somewhere to be shared, which needs the backend service for the database.

Does it need to save it? Doesn't it just need to generate a string to be copied? And then later, the ability to parse a pasted string to convert to a factory in the app? I'm a little concerned about the cost of maintaining 1000's of factories.

Maelstromeous:
> Doesn't it just need to generate a string to be copied?

If you're thinking akin to Factorio's blueprint string then that's a way of doing it, but that's acceptable for an in-game thing, whereas this is purely web based, and people don't like seeing gigantic links which only get worse with the more data there is in there.

I looked into compressing the string down, but there's a point where you just simply lose data. So, creating a link with a UUID (or in this case a nice enough mongodb object hash acting as the ID) you get best of both worlds, a unique link, which doesn't look like complete nonsense.

> I'm a little concerned about the cost of maintaining 1000's of factories.

Data is very cheap. I'm planning on implementing a purge system where if a link hasn't been visited for X days (30 say) it'll just delete the shares. Since the backend is a self hosted container, there's no worries about cloud costs / requests burning a hole in my pocket.

Maelstromeous:
https://github.com/user-attachments/assets/d3d884b2-a1d8-4de4-8e72-baec530cf294

Here's how the link generation experience is so far, next step for me is to make it so when the user visits the link it replaces their planner data with the data within the share.

This is what the data looks like stored within the database:

Screenshot 2024-11-19 at 00 02 04

samsmithnz:
> I looked into compressing the string down, but there's a point where you just simply lose data. So, creating a link with a UUID (or in this case a nice enough mongodb object hash acting as the ID) you get best of both worlds, a unique link, which doesn't look like complete nonsense.

What causes it to lose data? Is the Base64 string too long?

samsmithnz:
So therefore for testing scenarios - we will be accessing a string from the database with an ID?

Maelstromeous:
> What causes it to lose data? Is the Base64 string too long?
Simply the limits of compression. The more data there is, doesn't matter how much whitespace you strip out, even if it's base64 encoded it's still going to be quite long. 

I'm quite close with finishing the loading expereince for the share links, should have it done in the next hour.

Maelstromeous:
https://github.com/user-attachments/assets/11b23a14-382f-448c-b118-3e68a0e30562

@samsmithnz here's the full end to end experience of the sharing, let me know what you think!

Maelstromeous:
I've relaxed Sonar's quality gates, it won't bitch at us any more about code coverage or duplications so the statuses should pass from now on.

samsmithnz:
Nice! I'm curious about the thinking behind having to login? Is that to have access to the database?

samsmithnz:
If I have a factory, share it, and then decide it needs an update, does it create a new share link? Or overwrite the existing one?

Maelstromeous:
Currently a new factory will need to be made

Maelstromeous:
Brain dump before sleep time:
- [ ] Convert Dialog into a Toast, the user does not need to have to press a button to copy the link, it should be automatically copied and them informed it's done so for them to simply paste
- [ ] Need to hide the button when not on the planner view, currently it shows everywhere even outside of the planner e.g. /recipes
- [ ] Remove the need for an account, while my initial intention it would be an anti bot spam measure, it would likely do more harm than good. Could instead change it so after X amounts of shares from an IP they need to make an account perhaps. Also I could just be more paranoid about bots creating links than reasonable.

parsing/src/processor.ts
```diff
@@ -332,7 +332,7 @@ function getRecipes(
 // Create building object with the selected building and calculated power
             const building = {
                 name: selectedBuilding || '', // Use the first valid building, or empty string if none
-                power: String(powerPerBuilding || 0), // Use calculated power or 0
```
Maelstromeous, line 335: How the hell did this get in here? :O

parsing/src/interfaces/Recipe.ts
```diff
@@ -12,7 +12,7 @@ export interface Recipe {
   products: RecipeItem[];
   building: {
     name: string;
-    power: string;
```
Maelstromeous, line 15: Ehhh??

Maelstromeous:
Awaiting your full review @samsmithnz, I'm pretty happy with it now :)

backend/backend.ts
```diff
@@ -18,9 +21,21 @@ const PORT = 3001;
 // Setup Express
 // *************************************************
 
+// Configure rate limiter: maximum of 100 requests per 5 minutes (10 a minute)
```
samsmithnz, line 24: Do we have anything to track when the rate limit is reached?

backend/backend.ts
```diff
@@ -194,6 +211,65 @@ app.get('/load', authenticate, async (req: AuthenticatedRequest & TypedRequestBo
...
+    const { username } = req.user as jwt.JwtPayload & { username: string };
+    const factoryData = req.body;
+
+    console.log(`Creating share link for user ${username}`);
+
+    const shareId = await generateShareWords(3);
+
+    const shareData: ShareDataSchema = {
+      id: shareId,
+      data: JSON.stringify(factoryData),
```
samsmithnz, line 226: Perhaps an enhancement, but we could use `messagePack` here to compress the JSON.  
> According to tests and comparisons, MessagePack can compress data by up to 50% compared to JSON. This means a 10,000 character JSON string could potentially be reduced to around 5,000 characters when using MessagePack.

backend/backend.ts
Maelstromeous: It does this internally, I dare not question how, but I have tested it and it indeed does work.

There's actually now two layers of rate limiting:
1. Cloudflare rate limits site wide, restricted to 20reqs/10s
2. API app rate limits, using the two ones configured in this file.

I chose to be more stringent on the share one as there's possibility for abuse, whereas the save and load endpoints require user authentication, which could still be botted but at least it's more difficult to do so.

web/src/components/Sharing.vue
```diff
@@ -0,0 +1,70 @@
...
+      if (response.ok) {
+        const data: ShareDataCreationResponse = await response.json()
+        await navigator.clipboard.writeText(`${window.location.origin}/share/${data.shareId}`)
+        toast.value = true // Shows toast
+        creating.value = false
+      } else if (response.status === 429) {
+        alert('You are being rate limited. Stop spamming that button! Please wait some time before trying again.')
+      } else {
+        console.error('Creating share link failed failed:', response.body)
+        alert(`Failed to create share link. Please report this error to GitHub! "${response.body}"`)
```
samsmithnz, line 61:
```suggestion
        alert(`Failed to create share link. Please report this error on our GitHub site 'https://github.com/satisfactory-factories/application'! "${response.body}"`)
```

web/src/components/Sharing.vue
```diff
@@ -0,0 +1,70 @@
...
+      }
+    } catch (error) {
+      if (error instanceof Error) {
+        console.error('Error:', error)
+        alert(`Failed to create share link. Please report this error to GitHub! "${error}"`)
```
samsmithnz, line 66:
```suggestion
        alert(`Failed to create share link. Please report this error on our GitHub site 'https://github.com/satisfactory-factories/application'! "${error}"`)
```

samsmithnz (commented):
LGTM! 🚀  I added a couple of questions and suggestions, but nothing that should stop the merge.

samsmithnz (dismissed):

backend/backend.ts
Maelstromeous: I like the suggestion, I'll look into implementing it as it would be nice to cut down on any storage.

samsmithnz: created #89 to track

samsmithnz (approved):

backend/backend.ts
Maelstromeous: @samsmithnz Unfortunately it's using buffers, and with typescript buffers are a MASSIVE pain in the ass to get right and it be happy with it, so on this occasion I'm going to value my sanity over data storage, on a 12TB NAS pretty sure this will never be a problem :D

PR #96, opened 2024-11-19
Update processor.ts

### LINKED ISSUE #93: Fix missing parts
Hello. Whilst defining a Biofuel Factory I am unable to completely satisfy the factory because it does not know how to handle the ingredients for Biomass or Alien Protein. I have checked each, and the UNKNOWN PART is displaying for each.

image

samsmithnz:
fixes #93 - adding the raw materials for plants and animals - I don't know what the limit is, so put in a basic 1,000,000 as a placeholder. I was going to add slugs - but I don't see power shards in the recipes (just the synthetic alt recipe), so holding off on them for now.

### DIFF
--- a/parsing/src/processor.ts
+++ b/parsing/src/processor.ts
@@ -374,6 +374,8 @@ function getRawResources(data: any[]): { [key: string]: RawResource } {
         "Water": 9007199254740991,
         "Leaves": 1000000,
         "Wood": 1000000,
+        "Mycelia": 1000000,
+        
     }
 
     data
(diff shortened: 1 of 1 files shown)


Maelstromeous:
This likely isn't going to fix #93, the reason why the part is unknown is the `gameData.items[leaves]` entry does not exist because it is filtered out. Nothing "produces" leaves, therefore since it has no "building" it is omitted. We may need to add manual parts for that, or decide that we are not going to support them.

samsmithnz:
Bugger, I was trying to have a swing at this. I couldn't see anywhere in the code that included/filtered out other raw products - such as water and iron ore - just the raw material count. Where in the parser would this happen.

samsmithnz:
Can we add a line to manually add in leaves, wood, Mycelia, the four alien parts, and 3 power slugs? It would only be 10-12 lines, and solve it all.

samsmithnz:
Got it - I'll close this out!

PR #105, opened 2024-11-19
feat(web): #28 add default recipe selection

### LINKED ISSUE #28: Discussion about default recipes
I find myself clicking a lot to select all of the recipes for production - and I wonder if - by default - we should be selecting the default recipe, and changing to alt recipes - which at least in my games, is typically an exception. This would streamline the setting up of new factories in the tool. What do others think?

nils-keller-dev:
Closes #28

### DIFF
--- a/web/src/components/planner/PlannerFactoryProducts.vue
+++ b/web/src/components/planner/PlannerFactoryProducts.vue
@@ -276,6 +276,11 @@
       product.recipe = recipes[0].id
     }
 
+    const defaultRecipes = recipes.filter(recipe => !recipe.isAlternate)
+    if (defaultRecipes.length === 1) {
+      product.recipe = defaultRecipes[0].id
+    }
+
     updateFactory(factory)
   }
 


Maelstromeous:
This is actually quite an elegrant solution, very nice!

One small addition though, could you also do a check if "Recipe name === Product name" e.g. "Plastic" recipe for "Plastic" product to make it choose that, as that is preferable than just checking alts. 

Ideally it should check by exact recipe name first then by the alt check if there are no exact matches.

nils-keller-dev:
Sure thing, I pushed another commit

Maelstromeous:
Awesome stuff! Tested and it works really nicely! You've just saved many people many clicks, thank you! <3

nils-keller-dev:
Happy to help :)

Maelstromeous:
It really is appriciated! You're nailing some really nasty bugs, where I'm here trying not go insane with the amount of feedback I've had to absorb from Reddit, Discord DMs, YouTube, GitHub, it's a lot! 🥴

PR #111, opened 2024-11-20
fix(web): #76 fix +product button in Satisfaction

### LINKED ISSUE #76: The +product button in Satisfaction is not pre-filling the correct item in the new product
This was likely introduced by #60.

nils-keller-dev:
Closes #76 
These changes might be a bit controversial as I added functions to the `game-data-store.ts`.  
If you have other suggestions, let me know!

### DIFF
--- a/web/src/utils/factory-management/products.ts
+++ b/web/src/utils/factory-management/products.ts
@@ -10,16 +10,8 @@ export const addProductToFactory = (
     recipe?: string,
   }
 ) => {
-  // In order to allow multiple products of the same product but using different recipes to be added, we need to generate a unique ID using the product's ID as well as the recipe ID.
-
-  let productId = ''
-
-  if (options.id && options.recipe) {
-    productId = `${options.id}-${options.recipe}`
-  }
-
   factory.products.push({
-    id: productId,
+    id: options.id ?? '',
     amount: options.amount ?? 1,
     recipe: options.recipe ?? '',
     displayOrder: factory.products.length,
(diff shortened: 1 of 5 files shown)


web/src/utils/factory-management/products.ts
```diff
@@ -10,16 +10,8 @@ export const addProductToFactory = (
     recipe?: string,
   }
 ) => {
-  // In order to allow multiple products of the same product but using different recipes to be added, we need to generate a unique ID using the product's ID as well as the recipe ID.
-
-  let productId = ''
-
-  if (options.id && options.recipe) {
```
nils-keller-dev, line 17: This was never the case as none of the function calls passed a `recipe`.  
I removed it as it broke my implementation otherwise

Maelstromeous: Yeah this was the primary reason why the button was borked in the first place, I changed the expected data structure.

Maelstromeous:
Tests are failing fyi

Maelstromeous:
I like the refactor, recipe data is instringingly game data so yes absolutely makes sense to be in the store, then it is available anywhere in the code base. I do love the store concept!

Maelstromeous (changes requested):
Tests

nils-keller-dev:
I fixed the tests

Maelstromeous:
One day sonarqube will work again!

Maelstromeous:
Ok one thing I'd like you to test in the preview, please add a factory with duplicate products but different recipes. I believe using your implementation this might not work now, which would revert #9's fix.

Maelstromeous:
Nevermind, just quickly tested it myself with two iron plates, works fine! Approved!

Maelstromeous (approved):

PR #116, opened 2024-11-20
⚠️ MEATY PR: Exports Refactor, Factory Calculations relocation & Programmatic factory setups

### LINKED ISSUE #107: Imports are not being marked as invalid when dependant factory exports are no longer able to fulfil requests
I'm not sure why this is a problem. I've tried fidgeting with the numbers on things but it will not resolve. Factory "Copperhead" is in permanent red status. Hope i'm not missing something obvious and wasting your time.

https://satisfactory-factories.app/share/nutritious-adorable-businessperson

ROFL at the short-URL generator.

## Mael's edits

Repro Steps:

1. Create the following factories:
  - Copper Ingots @500/m
  - Wire (680) & Cables (300), import copper to meet demand
  - Stators (10) importing wire to meet demand (80)

This is a valid plan.

2. Increase **cables** @400. Wire internal demand will increase beyond production capabilities to 800.
3. BUG: Observe now how Stators has an invalid input. HOWEVER data still has the links remaining.
4. Increase stators to 20 and the wire import to 160
5. Wire satisfaction number has not changed. It should be 960 (800 internal demand + 160 from stators fac)

Here is the dependency data of the Wire fac, where Stators is still dependant, which is correct because the input still exists:

```json
"dependencies": {
    "requests": {
      "1086": [
        {
          "part": "Wire",
          "amount": 160
        }
      ]
    },
    "metrics": {
      "Wire": {
        "part": "Wire",
        "request": 160,
        "supply": 0,
        "isRequestSatisfied": false, <<< This is why the factory is red.
        "difference": -160
      }
    }
  },
  ```

So this bug of two parts:

## Exports display
Exports should be display requests REGARDLESS of surplus, right now if there is no surplus it does not display at all, meaning the `isRequestSatisfied` is causing the user frustration as the factory is red and they don't know why or how to fix it. They need to show IF there is a surplus OR there is an input for the product.

## Satisfaction calculation
It seems when we are in this state the dependency requirement portion of the satisfaction calculation utterly breaks. There is this warning in the console to suggest it, which is coming from `dependencies.ts`:
`calculateDependencyMetrics: Could not find a surplus for part Wire. Wire`

Also there's an additional bug with the imports display on the Stators factory where it is showing the numbers (presumably because now it's an invalid input factory). This should be resolved by #113.

### PR DESCRIPTION
**Pull request by Maelstromeous**
Closes #107 - If products have an internal deficit they drop off the exports list as they no longer have a surplus. This PR aims to refactor the concept of Exports into a more usable form, rather than a mish-mash between Vue and Data combined, leaning more towards the data side and letting the View do less calculation work.

This is a four part PR:

## Exports refactor

This PR refactors the data generation of exports (was called surplus) and disregards product surplus as the sole filter of this view. It now loops all products and checks if they have a surplus OR demands of the product from other factories.

The view has been heavily refactored to remove the logic that it once had, instead opting to simply use the data, like a view should.

## Factory Calculations relocation

It no longer makes much sense to keep the `calculateFactory` and it's sibling `calculateFactories` function within `Planner.vue`. We can proxy the call to `caluclateFactories` from the vue and inject the game data that way at runtime. Otherwise, we can pull in the game data via an import constant.

This makes testing the code A LOT easier, and also removes a hard coded dependency of `Planner.vue` in calculations of the factories. It should have been abstracted long ago.

## Programmatic factory setup
A weakness we had is our VERY large demo factories object needed to be maintained with changes to the overall data structure, or the very first (**and very important**) impression of the site would be broken, and we wouldn't know it. I have now converted this to be programmatically created, so the maintenence of the object is eliminated.

Also this means we can create "setups" for factories for debugging and testing purposes. 

Ideally when the site is in debug mode we need to be able to press a button and load the setups at will.

## `factories.value` refactor

So changing `factory.surplus` into `factroy.exports` broke the site locally for me on old data. This WILL break user's data as well. Therefore, we are enforcing the use of the `appStore` within a data service provider pattern, calling a getter rather than using the naughty way of getting the value directly. Using this getter we are able to inject patches into the data, to ensure we don't break people's factories going forward.

### CODE REVIEW

**Comment by Maelstromeous** [comment]
Expanding this PR to re-write the entire data on exports, it needs to be calculated better rather than having the template to do it all.

File: web/src/interfaces/planner/FactoryInterface.ts [hunk 1]
```diff
@@ -92,9 +95,9 @@ export interface Factory {
   buildingRequirements: { [key: string]: BuildingRequirement };
   requirementsSatisfied: boolean;
   totalPower: number;
-  surplus: { [key: string]: FactorySurplusItem };
-  dependencies: FactoryDependency;
+  exports: { [key: string]: FactoryExportItem };
   exportCalculator: { [key: string]: ExportCalculatorSettings };
+  dependencies: FactoryDependency;
```
Maelstromeous, line 100: Not new just moved

web/src/utils/factory-management/dependencies.ts
```diff
@@ -97,15 +97,11 @@ export const calculateDependencyMetrics = (factory: Factory) => {
...
-      if (!factory.surplus[part]) {
-        console.warn(`calculateDependencyMetrics: Could not find a surplus for part ${part}.`, part)
-      }
-
       if (!metrics[part]) {
         metrics[part] = {
           part,
           request: 0,
-          supply: factory.surplus[part]?.amount ?? 0,
+          supply: factory.parts[part].amountRemaining ?? 0,
```
Maelstromeous, line 104: This removes a circular dependency between export data and dependency metrics data, by using the part data as the source, which both of them use anyway.

web/src/utils/factory-management/products.spec.ts
Maelstromeous, line 1: No substantial changes, just gameData reference.

web/src/utils/factory-management/satisfaction.spec.ts
Maelstromeous, line 1: No substantial changes, just gameData reference.

web/src/utils/factory-management/supply.spec.ts
Maelstromeous, line 1: No substantial changes, just gameData reference.

Maelstromeous:
Just want to highlight how easy it is to create rapid sets of test cases with the new test factory template concept:

```typescript
it('should calculate exports correctly with confirmed bugged factory', () => {
    const factories = internalProductionDeficit()

    calculateFactories(factories, gameData)

    const wireFac = findFacByName('Wire', factories)
    expect(wireFac.exports.Wire.surplus).toBe(0)
    expect(wireFac.exports.Wire.demands).toBe(170)
  })
  ```
  
  That pulls in a fully dynamically generated factory and runs a very specific test against it.

samsmithnz (approved):

PR #123, opened 2024-11-20
fix(web): fix rounding errors in product calculation

### LINKED ISSUE #102: Rounding errors in product calculations
We have issues where under certain circumstances a 0.00000000001 less of a product is being made.

All product values need to be rounded to the nearest 0.xxx of an integer, and stored as such to prevent these extremes of ever-so-slightly-off. 

The fallout of this is Satisfaction is showing 2000/2000 as unsatisfied and users are complaining that this is occuring.

Trimming and fixing productions appear to introduce the floats, not quite sure why it would but it is. There must be some wonky math going on, or the nature of working with floats, one of the two.

nils-keller-dev:
Closes #102

### DIFF
--- a/web/src/utils/factory-management/products.ts
+++ b/web/src/utils/factory-management/products.ts
@@ -64,10 +64,8 @@ export const calculateProducts = (factory: Factory, gameData: DataInterface) => 
 
       // === Now we need to calculate the parts required per min to make the product ===
       const productIngredientRatio = product.amount / recipe.products[0].perMin // This formular should have no rounding - the decimal points are important here for the floating point calculations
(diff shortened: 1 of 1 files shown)


samsmithnz (approved):

Maelstromeous:
In my rudimentary testing this appears to work, would have been nice if we had tests specificly around the issue but eh

Maelstromeous (approved):

samsmithnz:
The tests definitely cover this! When I fixed the rounding last time, I reviewed them all and ensured they were 100% correct - I'm confident this is covered.

PR #124, opened 2024-11-20
Add Factory tabs

### LINKED ISSUE #109: Add Planner Pages
It is becoming increasingly apparent that we need some means to create seperate pages of plans. There are a number of drivers for this:

1. Loading a shared link means it will blow your plan away. People will not want to share links knowing that the recipient will have to remove their data, or a hacky workaround where they have to "share" their data to effectively export it.
2. Debugging other people's plans in production would be MUCH easier if we could load the plan in a new tab, then delete the tab when we have finished our diagnosis.
3. This also gives users the ability to collaborate with others on a planner page, which to me is a prelude requirement for #33. 

The implementation of this should be relatively trivial. Rather than using one big "factories" object, we can simply create a new page and use the factory object in there. There would need to be an object that controls tabs, and there will be some UI work in order to manage them etc, but the data contained within should be reletively untouched.

For now, I don't think we should add the ability for users to do imports or links between pages, that would be a bit mental, BUT it could be done, just not for an initial first pass.

nils-keller-dev:
Closes #109 

I did not add the Icon picker as the PR got a little big already

Here are my key changes:
* When opening the site the user starts with one "Default" tab
* There should be no data loss to users when this version gets deployed as the "factories" localStorage entry is used as fallback when there is no new "factoryTabs" entry
* You can add, delete, share and rename tabs
* Each tab contains its own set of factories
* When opening a share link, it gets added as a new tab
 -> share links that were created before this change will break, should I fix this?

Does it make sense to explain this functionality in the welcome modal? I'll let you decide

Please make sure to test this well before deploying, as this may break persistent data for users!

### DIFF
--- a/web/src/components/ShareButton.vue
+++ b/web/src/components/ShareButton.vue
@@ -1,14 +1,11 @@
 <template>
   <v-btn
-    class="ml-2"
-    color="blue"
-    :disabled="creating"
+    color="blue rounded"
+    icon="fas fa-share-alt"
+    size="small"
     variant="flat"
     @click="createShareLink"
-  >
-    <i class="fas fa-share-alt" /><span class="ml-2">Share Plan</span>
-  </v-btn>
-
+  />
   <v-snackbar v-model="toast" color="success" top>
     Link copied to clipboard!
   </v-snackbar>
--- a/web/src/components/ShareButton.vue
+++ b/web/src/components/ShareButton.vue
@@ -15,27 +15,27 @@
   import { config } from '@/config/config'
   import { storeToRefs } from 'pinia'
   import { useAppStore } from '@/stores/app-store'
-  import { Factory } from '@/interfaces/planner/FactoryInterface'
+  import { FactoryTab } from '@/interfaces/planner/FactoryInterface'
   import { ShareDataCreationResponse } from '@/interfaces/ShareDataInterface'
 
   // Get user auth stuff from the app store
   const appStore = useAppStore()
-  const { token, factories } = storeToRefs(appStore)
+  const { token, currentFactoryTab } = storeToRefs(appStore)
 
   const apiUrl = config.apiUrl
   const toast = ref(false)
   const creating = ref(false)
 
   const createShareLink = async () => {
-    if (!factories.value || factories.value.length === 0) {
+    if (!currentFactoryTab.value.factories || currentFactoryTab.value.factories.length === 0) {
       alert('No factory data to share!')
       return
     }
 
-    await handleCreation(factories.value)
+    await handleCreation(currentFactoryTab.value)
   }
 
-  const handleCreation = async (factoryData: Factory[]) => {
+  const handleCreation = async (factoryTabData: FactoryTab) => {
     creating.value = true
     try {
       const response = await fetch(`${apiUrl}/share`, {
@@ -44,7 +44,7 @@
           'Content-Type': 'application/json',
           Authorization: `Bearer ${token.value}`,
         },
-        body: JSON.stringify(factoryData),
+        body: JSON.stringify(factoryTabData),
       })
       if (response.ok) {
         const data: ShareDataCreationResponse = await response.json()
--- a/web/src/components/ShareButton.vue
+++ b/web/src/components/ShareButton.vue
@@ -1,6 +1,7 @@
 <template>
   <v-btn
     color="blue rounded"
+    :disabled="creating"
     icon="fas fa-share-alt"
     size="small"
     variant="flat"
--- a/web/src/components/TabNavigation.vue
+++ b/web/src/components/TabNavigation.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <v-toolbar class="border-t-md" dark density="compact">
+    <div class="border-t-md d-flex bg-grey-darken-3 align-center" density="compact">
       <v-tabs
         v-model="appStore.currentFactoryTabIndex"
         color="deep-orange"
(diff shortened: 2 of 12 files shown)


Maelstromeous:
> When opening a share link, it gets added as a new tab
-> share links that were created before this change will break, should I fix this?
Ah there may be a slight misunderstanding here perhaps, the share links should only be sharing the currently viewed tab.

OR

We make it so that the user can select what tabs they can choose to share. This is likely out of scope of this PR now though, so lets make it so that the share shares the currently viewed plan. Reason for this is numerous, but mainly so that we don't blow away the user's full set of plans with the share, which was the impotus for this feature in the first place.

Maelstromeous:
In the interests of saving some vertical space, could we try some of this?

Screenshot 2024-11-20 at 23 57 40

Namely move the share and bin buttons up, and then add a pencil next to the title to enable the user to edit the title and remove the textbox?

web/src/pages/share/[id].vue
```diff
@@ -47,22 +30,18 @@
 
   const loadShareData = async () => {
     const shareId = (route.params as { id: string }).id
+    if (!shareId) return
```
Comment (Maelstromeous) [inline web/src/pages/share/[id].vue:33]: Should really redirect back to the planner
```suggestion
    if (!shareId) router.push('/')
```

web/src/pages/share/[id].vue
```diff
@@ -76,9 +55,8 @@
       const data: ShareDataReturnResponse = await response.json()
       if (response.ok) {
         console.log('Data:', data)
-        loading.value = false
 
-        if (data.data.length === 0) {
+        if (!data.data) {
```
Comment (Maelstromeous) [inline web/src/pages/share/[id].vue:59]: Double checking, will this work on an empty array?

web/src/stores/app-store.ts
```diff
@@ -1,33 +1,46 @@
...
-  // Getters
-  const getFactory = (id: number) => {
-    return factories.value.find(factory => factory.id === id)
-  }
-  const getFactories = () => {
-    return factories.value
-  }
-  const getLastEdit = () => {
-    return lastEdit.value
-  }
```
Maelstromeous, line 29: Please don't delete these, as they are partially used by #116 and getLastEdit will eventually get used by Auth for synchronisation checking.

web/src/stores/app-store.ts
```diff
@@ -80,16 +93,41 @@ export const useAppStore = defineStore('app', () => {
...
+    currentFactoryTabIndex,
+    factoryTabs,
     factories,
     loggedInUser,
     token,
     lastSave,
     lastEdit,
-    getFactory,
-    getFactories,
-    getLastEdit,
```
Maelstromeous, line 92: Please undo

Maelstromeous (changes requested):
Excellent stuff! Left a few things to look into

Maelstromeous:
> Does it make sense to explain this functionality in the welcome modal? I'll let you decide

Probably not needed, looks self explanatory to me!

web/src/components/TabNavigation.vue
```diff
@@ -0,0 +1,67 @@
...
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+  import { useAppStore } from '@/stores/app-store'
+
+  const appStore = useAppStore()
+
+  const onEnterTabName = (event: KeyboardEvent) => {
```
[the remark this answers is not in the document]
nils-keller-dev: I deliberately used `v-model.lazy` so it only updates on `change` and not on `input`.
I did this so the localStorage does not get written every time the user presses a key, because that felt like a bad idea.

web/src/components/ShareButton.vue
```diff
@@ -1,14 +1,11 @@
 <template>
   <v-btn
-    class="ml-2"
-    color="blue"
-    :disabled="creating"
```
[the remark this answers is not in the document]
nils-keller-dev: Sure, good catch

web/src/components/graph/Graph.vue
```diff
@@ -75,6 +75,10 @@
     console.log('Initial nodes:', nodes.value)
     console.log('Initial edges:', edges.value)
   })
+
+  watch(factories, () => {
+    initializeGraph()
```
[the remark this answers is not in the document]
nils-keller-dev: This was to make the tab switching work in the graph view

web/src/pages/share/[id].vue
Comment (nils-keller-dev) [reply web/src/pages/share/[id].vue:59]: No it won't, but `data` will be an object instead of an array from now on

web/src/stores/app-store.ts
nils-keller-dev: Got it!

web/src/stores/app-store.ts
nils-keller-dev: 👍

nils-keller-dev:
> Ah there may be a slight misunderstanding here perhaps, the share links should only be sharing the currently viewed tab, so when users visit a share it should simply just make a new tab and navigate them to it.
> Edit: Looks like you're doing this so it should work?

Yes this is the way I implemented it. However if a share link is created before this PR gets deployed, it will not work once it has been deployed since it uses a different data structure

nils-keller-dev:
> In the interests of saving some vertical space, could we try some of this?
> 
> Screenshot 2024-11-20 at 23 57 40
> Namely move the share and bin buttons up, and then add a pencil next to the title to enable the user to edit the title and remove the textbox?

I will give this a try shortly

Maelstromeous:
> Yes this is the way I implemented it. However if a share link is created before this PR gets deployed, it will not work once it has been deployed since it uses a different data structure

It could do. So from what I understand of your new code you're just copying the currently open tab, so, therefore, it should be possible to create a new tab and inject the data from the share link into it. The share link is just a copy of the `factories` data in the old way of doing it, it should be entirely possible just to put that a level below `factoryTabs` in a new tab and replace it's data with the share?

web/src/components/TabNavigation.vue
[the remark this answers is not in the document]
Maelstromeous: Ah i see, I've never encountered any issues with localStorage not updating the state, every time a factory value gets change it's written to localStorage as there's a watcher on the factory data.

nils-keller-dev:
I will close this PR and all future work will be done in https://github.com/satisfactory-factories/application/pull/129

PR #129, opened 2024-11-21
Add Factory tabs

### LINKED ISSUE #109: Add Planner Pages

nils-keller-dev:
Closes #109 
This is the same PR as this one: https://github.com/satisfactory-factories/application/pull/124
But I moved the branch from my fork to this repo


nils-keller-dev:
@Maelstromeous have a look at my last 3 commits, everything should be resolved now

Maelstromeous (approved):
Looks awesome! Well done!

PR #137, opened 2024-11-21
feat(workflow): Remove 'Needs Triage' label from issues after adding P1 label

samsmithnz:
This pull request includes a change to the `.github/workflows/issue-management.yml` file to improve the issue management workflow by removing the 'Needs Triage' label from issues.

### DIFF
--- a/.github/workflows/issue-management.yml
+++ b/.github/workflows/issue-management.yml
@@ -45,6 +45,12 @@ jobs:
                 milestone: milestone.number
               });
             }
+            await github.rest.issues.removeLabel({
+              issue_number: issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


Maelstromeous (approved):

.github/workflows/issue-management.yml
```diff
@@ -45,6 +45,12 @@ jobs:
                 milestone: milestone.number
               });
             }
+            await github.rest.issues.removeLabel({
```
Maelstromeous, line 48: @samsmithnz This should probably be outside of the P1 IF condition, we want to do this on **any** issue that recieves labels really.

Maelstromeous: S33 https://github.com/satisfactory-factories/application/issues/138 as an example

PR #140, opened 2024-11-22
Where is Wallys nuclear waste?

### LINKED ISSUE #133: UNKNOWN PART: PlutoniumWaste, NuclearWaste
Trying to produce "Ficsonium", results in a  *requires* "UNKNOWN PART PlutoniumWaste"
Similary "Plutonium Pellet" and "Non-Fissile Uranium" results in a **requires** "UNKNOWN PART NuclearWaste!"

samsmithnz:
fixes #133 to add Uranium Waste and Plutonium Waste items and recipes to produce it. It is odd, as a power generating building creates it as a byproduct, and hence, doesn't consume any power. But this should now work in the planner.

<HR>

This pull request includes updates to the handling of nuclear waste and power plant data in the game data files and processing scripts. The changes introduce new items, recipes, and manual adjustments for nuclear waste products.


samsmithnz:
Ehhhhh. Recipes didn't come through. Going to chase and see what happened to them

Maelstromeous:
Aww I rather liked that title :D

Maelstromeous:
So while the recipe is now there, the Items are missing.

Also I'm not sure this is the right approach. Player's dont want to inrinsingsly "produce" waste via a recipe, it'll just appear as an issue by burning nuclear fuel.

Take a look at the wiki for example, the only usage of PlutoniumWaste is in a single recipe for Ficsonium. So I'd be more inclined that we add the "Burn fuel" recipe which produces nuclear waste, rather than trying to produce the item itself.

This unfortunately stems into the realm of "Power Recipes" which are used for consumer destinations e.g. #7, which we want to steer clear of for now in this PR.

To resolve #133, we need to have the Item data fixed, which unfortunately this PR doesn't yet do. Apologies if it's still WIP, it's not indicated to me it's a draft PR.

samsmithnz:
I can fix one of your problems. The items should be there. Will go back and look again. This is why code reviewers work so well. They see what is blind to others. I was too focused on getting the recipes to appear I think!

Maelstromeous:
Yep, they belong under here: https://github.com/satisfactory-factories/application/pull/140/files#diff-8c0c202e320dc48bd8ebc0ce35480877f8a06ade9508a321e2e0979acffbfed7R17

samsmithnz:
Progress! Nuclear Waste now exists. But not Nuclear Fuel Rods do not. Pulling the string! 

image

samsmithnz:
Progress! I can't get the Nuclear Power Plant image to show up, but this is an improvement. The recipes are showing now! 

image

samsmithnz:
I think this addresses most of the concerns above. If I've missed anything, let me know and I will address.

parsing/src/processor.ts
```diff
@@ -93,8 +93,27 @@ function getItems(data: any[]): PartDataInterface {
...
+                    name: "Plutonium Waste",
+                    stackSize: 500, //SS_HUGE
+                    isFluid: false,
+                    isFicsmas: false,
+                };
+            }
+
             // Ensures it's a recipe, we only care about items that are produced within a recipe.
-            if (!entry.mProducedIn) return;
+            if (!entry.mProducedIn) return;
```
Maelstromeous, line 116: Wonky spacing

Maelstromeous (commented):
Will do a full test tomorrow

parsing/src/processor.ts
samsmithnz: We should consider adding a linter to break the build when this stuff is added.

Maelstromeous: We could yes

parsing/src/processor.ts
```diff
@@ -180,6 +199,8 @@ function getProducingBuildings(data: any[]): string[] {
                 if (producedInBuildings) {
                     producedInBuildings.forEach((buildingName: string) => producingBuildingsSet.add(buildingName));
                 }
+            } else if (entry.ClassName === "Desc_NuclearWaste_C") {
```
Maelstromeous, line 202: Any reason why we can't just add this in at the end of the processing rather than running the conditional against every entry?

samsmithnz: Definitely focused on a 'build, then optimize' methodology. The code runs in less than a second. Now that it works, I want to focus on tests and optimizing. In Particular, I think some of these if's can be removed with some tests and optimization.

Maelstromeous: Aight I'll let this one go for now then, you're right premature optimisation is the root of all evil as they say

Maelstromeous (approved):

PR #141, opened 2024-11-22
feat(web): Added new FixProduction button to Products (and made them green)

### LINKED ISSUE #27: Add "Fix Production" button near the Product
Hi, it would be great if you can add the "Fix Production" button
image
near the Product
image

It would be like what you already did with the Imports, that it says "Trim" or "Satisfy if the quantity is not enough or too much.
The feature already exists, but is not added in the Product rows

AndrewThomasWill:
- Added new 'Fix Production' button to the Product (tried to size it the best I could to match)
- Moved logic for fixing the part to 'Factory' vue level
- Updated both of the buttons to 'green' (as requested by Maelstrom)

Closes #27

### DIFF
--- a/web/src/components/planner/Planner.vue
+++ b/web/src/components/planner/Planner.vue
@@ -83,7 +83,6 @@
     ByProductItem,
     Factory,
     FactoryItem,
-    PartMetrics,
     WorldRawResource,
   } from '@/interfaces/planner/FactoryInterface'
   import { DataInterface } from '@/interfaces/DataInterface'
@@ -256,11 +255,6 @@
     return product ?? byProduct ?? undefined
   }
 
-  const getPart = (factory: Factory, partId: string): PartMetrics | undefined => {
-    const part = factory.parts[partId]
-    return part ?? undefined
-  }
-
   const getBuildingDisplayName = (building: string) => {
     const buildingFriendly = new Map<string, string>([
       ['assemblermk1', 'Assembler'],
@@ -372,7 +366,6 @@
   provide('moveFactory', moveFactory)
   provide('isItemRawResource', isItemRawResource)
   provide('getProduct', getProduct)
-  provide('getPart', getPart)
 
   // Grab from local storage if the user has already dismissed this popup
   // If they have, don't show it again.
--- a/web/src/components/planner/Planner.vue
+++ b/web/src/components/planner/Planner.vue
@@ -83,6 +83,7 @@
     ByProductItem,
     Factory,
     FactoryItem,
+    PartMetrics,
     WorldRawResource,
   } from '@/interfaces/planner/FactoryInterface'
   import { DataInterface } from '@/interfaces/DataInterface'
@@ -255,6 +256,11 @@
     return product ?? byProduct ?? undefined
   }
 
+  const getPart = (factory: Factory, partId: string): PartMetrics | undefined => {
+    const part = factory.parts[partId]
+    return part ?? undefined
+  }
+
   const getBuildingDisplayName = (building: string) => {
     const buildingFriendly = new Map<string, string>([
       ['assemblermk1', 'Assembler'],
@@ -366,6 +372,7 @@
   provide('moveFactory', moveFactory)
   provide('isItemRawResource', isItemRawResource)
   provide('getProduct', getProduct)
+  provide('getPart', getPart)
 
   // Grab from local storage if the user has already dismissed this popup
   // If they have, don't show it again.
--- a/web/src/components/planner/PlannerFactory.vue
+++ b/web/src/components/planner/PlannerFactory.vue
@@ -164,7 +164,7 @@
 
 <script setup lang="ts">
   import { defineProps, inject } from 'vue'
-  import { Factory } from '@/interfaces/planner/FactoryInterface'
+  import { Factory, FactoryItem } from '@/interfaces/planner/FactoryInterface'
   import { DataInterface } from '@/interfaces/DataInterface'
   import { getPartDisplayName } from '@/utils/helpers'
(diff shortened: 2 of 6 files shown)


web/src/components/planner/Planner.vue
```diff
@@ -255,6 +256,11 @@
     return product ?? byProduct ?? undefined
   }
 
+  const getPart = (factory: Factory, partId: string): PartMetrics | undefined => {
```
Maelstromeous, line 259: This is fairly pointless, as you can just reference the part object out of the factory directly. It offers nothing of real value.

web/src/components/planner/Planner.vue
```diff
@@ -366,6 +372,7 @@
   provide('moveFactory', moveFactory)
   provide('isItemRawResource', isItemRawResource)
   provide('getProduct', getProduct)
+  provide('getPart', getPart)
```
Maelstromeous, line 375: Especially when we have to set up a provider for it

web/src/components/planner/PlannerFactoryProducts.vue
```diff
@@ -64,6 +64,13 @@
           variant="outlined"
           @input="updateFactory(factory)"
         />
+        <v-btn
+          v-if="!getPart(factory, product.id).satisfied"
```
Maelstromeous, line 68:
```suggestion
          v-if="!factory.part[product.id]?.satisfied"
```
You can use a null chain operator here to say if the object doesn't exist don't bother trying to get the attribute just return undefined. 

Double check this though.

Maelstromeous (changes requested):

web/src/components/planner/Planner.vue
AndrewThomasWill: Sure, I can remove this. I was just following the pattern of getting a Product since they are similar. But I'm all for deleting it 👍

web/src/components/planner/PlannerFactoryProducts.vue
AndrewThomasWill: Yeah, I can simplify it that way

web/src/components/planner/Planner.vue
AndrewThomasWill: Removed in latest commit

web/src/components/planner/PlannerFactoryProducts.vue
AndrewThomasWill: Updated in latest commit

Maelstromeous:
Screenshot 2024-11-22 at 15 20 32

Ideally if an export has a production shortage we should be showing a Fix Production button in that scenario as well.

This is fairly easy to calculate, simply a `if (factory.exports[part].difference < 0`.

AndrewThomasWill:
> Screenshot 2024-11-22 at 15 20 32
> Ideally if an export has a production shortage we should be showing a Fix Production button in that scenario as well.
> 
> This is fairly easy to calculate, simply a `if (factory.exports[part].difference < 0`.

Unless you just added this to prod, it seems to have already been there 😅

image

Maelstromeous:
> > Ideally if an export has a production shortage we should be showing a Fix Production button in that scenario as well.
> > This is fairly easy to calculate, simply a `if (factory.exports[part].difference < 0`.
> 
> Unless you just added this to prod, it seems to have already been there 😅
> 
> image

No, on the product. If the export is starved the product should show Fix Production.

AndrewThomasWill:
Imports Difference
Production Difference
Satisfaction Difference
Export Difference

Everything should be cleaned up and worded correctly now

-- Move 'getRequestMetricsForFactoryByPart()' and 'fixShortage()' one level higher to 'PlannerFactory' to be reused in other child components
-- Update 'fixShortage()' to call getting the product internally (instead of making child call getProduct)
-- Update 'PlannerFactorySatisfaction' to use the 'fixProduction' in 'PlannerFactory' instead of the local one
-- Rename some local/function variables to be more accurate

web/src/components/planner/PlannerFactoryExports.vue
```diff
@@ -62,8 +62,8 @@
                   color="green"
                   size="small"
                   variant="flat"
-                  @click="fixShortage(factory, getProduct(factory, exportItem.productId))"
-                >Fix production</v-btn>
+                  @click="fixExport(factory, exportItem.productId)"
+                >Fix Export</v-btn>
```
Maelstromeous, line 66:
```suggestion
                >Fix Production</v-btn>
```

web/src/components/planner/PlannerFactoryExports.vue
```diff
@@ -62,8 +62,8 @@
                   color="green"
                   size="small"
                   variant="flat"
-                  @click="fixShortage(factory, getProduct(factory, exportItem.productId))"
-                >Fix production</v-btn>
+                  @click="fixExport(factory, exportItem.productId)"
+                >Fix Production</v-btn>
```
Maelstromeous, line 66: Changed this as we're actually fixing the Production, rather than the export. Otherwise the user might get a bit confused.

web/src/components/planner/PlannerFactory.vue
```diff
@@ -215,6 +216,57 @@
...
+    if (!product) {
+      console.error(`Could not find product for ${productId} to fix!`)
+      return
+    }
+
+    const metric = getRequestMetricsForFactoryByPart(factory, product.id)
+
+    if (!metric) {
+      console.error(`Could not get request metric to fix shortage for ${product.id}`)
+      return
```
Maelstromeous, line 247: This doesn't actually do anything?

Maelstromeous (approved):
Looks great! Thank you! 🚀

PR #145, opened 2024-11-22
Authentication and sync refactoring

### LINKED ISSUE #136: Refactor: Break up Auth functionality in appStore into an authStore
`appStore` is doing WAY too much. We have issues with our auth and syncing systems and the `appStore` is too unwiedly to properly debug. Time to break it up.

Also `Auth.vue` is doing too much as well with authentication. It is a view, it is not meant to be a authentication manager.

Prelude to work on #135 #130, #77 as we need to completely re-organise the current mess that the auth and sync features are.

Maelstromeous:
Closes #136

### DIFF
--- a/web/src/stores/app-store.ts
+++ b/web/src/stores/app-store.ts
@@ -6,7 +6,6 @@ import { calculateFactories } from '@/utils/factory-management/factory'
 import { useGameDataStore } from '@/stores/game-data-store'
 
 export const useAppStore = defineStore('app', () => {
-  // Define state using Composition API
   const factoryTabs = ref<FactoryTab[]>(JSON.parse(localStorage.getItem('factoryTabs') ?? '[]') as FactoryTab[])
 
   if (factoryTabs.value.length === 0) {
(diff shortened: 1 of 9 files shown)


web/src/stores/app-store.ts
```diff
@@ -32,19 +31,31 @@ export const useAppStore = defineStore('app', () => {
...
-  const loggedInUser = ref<string>(localStorage.getItem('loggedInUser') ?? '')
-  const token = ref<string>(localStorage.getItem('token') ?? '')
   const lastSave = ref<Date>(new Date(localStorage.getItem('lastSave') ?? ''))
   const lastEdit = ref<Date>(new Date(localStorage.getItem('lastEdit') ?? ''))
   const isDebugMode = ref<boolean>(false)
   const gameDataStore = useGameDataStore()
 
   // Watch the factories array for changes
-  watch(factoryTabs, () => {
+  watch(factoryTabs.value, () => {
```
Maelstromeous, line 40: Has no change, just consistency.

nils-keller-dev:
I could not get vue component tests to work with your config, so I added vitest and a example test file instead. If you don't agree with that decision, you can of course revert my commit

PR #147, opened 2024-11-22
Fix linked cloned factories

### LINKED ISSUE #142: Cloned Factories are somehow linked - updating one, updates the other
https://satisfactory-factories.app/share/shaggy-faint-exabyte

If you load this and open Space Elevator Part 1 or Space Elevator Part 2 - and change an item, it changes in both factories.

nils-keller-dev:
Closes #142 

I was going to add a test, but the function is currently in a .vue file and I think it is fine there. However currently we cannot test vue files without reconfiguring jest and adding packages.
Should we tackle that? Also I would recommend using vitest, as it uses the same api and has better vue support

### DIFF
--- a/web/src/components/planner/Planner.vue
+++ b/web/src/components/planner/Planner.vue
@@ -202,10 +202,10 @@
   }
 
   const copyFactory = (originalFactory: Factory) => {
-    // Make a shallow copy of the factory with a new ID
+    // Make a deep copy of the factory with a new ID
     const newId = Math.floor(Math.random() * 10000)
     const newFactory = {
-      ...originalFactory,
+      ...JSON.parse(JSON.stringify(originalFactory)),
       id: newId,
       name: `${originalFactory.name} (copy)`,
       displayOrder: originalFactory.displayOrder + 1,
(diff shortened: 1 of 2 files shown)


Maelstromeous:
>  However currently we cannot test vue files without reconfiguring jest and adding packages.
Should we tackle that? Also I would recommend using vitest, as it uses the same api and has better vue support

Sure if vitest enables us to more easily test vue files I'm all for it. I was looking into this myself but then decided for the code I'm working on just to proxy it to a .ts file and test that instead.

I think we're in desperate need of testing the TS within .vue files as there are key bits of logic that aren't being looked after and break occasionally.

Maelstromeous:
Out of interest how did you validate that this fixed the original issue?

Maelstromeous:
Screenshot 2024-11-22 at 22 26 27

Seems to be working anyway, copied the computers factory and changed some values

Maelstromeous:
After understanding the repro steps a bit better, can confirm it's fixed:

Screenshot 2024-11-22 at 22 33 12

Screenshot 2024-11-22 at 22 33 46

Maelstromeous (approved):

PR #148, opened 2024-11-22
Fix broken exports section

### LINKED ISSUE #134: Exports: If there is an empty product in the factory, exports vanishes
https://satisfactory-factories.app/share/astonishing-steep-airline

Observe exports are not rendering, remove empty product, they render again.

nils-keller-dev:
Closes #134 

The error occurred when an item gets removed from the input
The Buildings & Power section also got broken when that happened, I fixed that too


Maelstromeous (approved):
Nice one! :D

PR #161, opened 2024-11-24
Adding test that would have helped us find a missing product sooner

samsmithnz:
Simple test to prevent bugs if new recipes are added and we are missing the products in the parts list too. Verified it failed tests when I removed a part and ran the new test.

Unfortunately, we can't do this with ingredients - as there are a number of machines that produce recipes with no ingredients - converters in particular. 

<HR>

This pull request includes changes to the `parsing/tests/parsing.spec.ts` file to improve the test coverage and ensure that recipes have products.


parsing/tests/parsing.spec.ts
```diff
@@ -75,6 +75,10 @@ describe('common', () => {
                 for (const product of recipe.products) {
                     parts.add(product.part);
                 }
+                if (recipe.products.length === 0) {
+                    console.log('No products for recipe:', recipe.id);
+                    expect('Recipe ' + recipe.id + ' has no products').toBe('');
```
Maelstromeous, line 80: Sorry this one doesn't match a ton of sense, you've got an `expect` which is asserting a string to be `''`. It's not actually testing anything?

Maelstromeous (changes requested):

parsing/tests/parsing.spec.ts
Maelstromeous: All it does is expects something in the case of recipe.products.length is 0.

It would be better as:

`expect(recipe.products.length).not.toBe(0)`

samsmithnz: It's more an assert. The thinking here is if there are no products, it throws an error with the description of the recipe that has the issue, so that it's easier to resolve. If I just  `expect(recipe.products.length).not.toBe(0)`, I have no idea what recipe failed and I have to start a whole new line of debugging. 

Does that make sense, or still sound crazy?

Maelstromeous: If you change `console.log` to `console.error` and **not** use `--silent` on the test run it'll show up the console in the output.

samsmithnz: That is another solution. I left the `--silent` on because you started with it (I usually prefer verbose for my tests). I'm flexible here. Whatever you think is best. :)

Maelstromeous: Well what we can do is make it fail, and if it's failing we can run the tests without --silent and see why, best of both words :)

samsmithnz: I'm not clear what you need to close this out. Please be more specific and I'll do it! :)

Maelstromeous: > `expect(recipe.products.length).not.toBe(0)`

This, we need the test to fail, you can keep the logging but do it as console.errors

Maelstromeous (approved):

PR #163, opened 2024-11-24
refactor(web): #136 improve store tests

nils-keller-dev:
@Maelstromeous I made some improvements to the store tests, mainly removing the need fr overrides


Maelstromeous:
Integrated suggestions into #136

PR #175, opened 2024-11-25
Fixes bug to show details for part with no ingredients

### LINKED ISSUE #171: Converters are missing building and power values
See `Excited Photonic Matter` item in factory below. At an initial guess, it may be because there are no ingredients for this recipe - it's one of the first recipes without ingredients.

Image

samsmithnz:
fixes #171 

Previously logic was only looking if there were ingredients before showing/hiding the row, but some parts (in phase 5), do not have ingredients - just power and a product.

<HR>

This pull request includes an important change to the `PlannerFactoryProducts.vue` file to improve the display logic for product requirements.

### DIFF
--- a/web/src/components/planner/PlannerFactoryProducts.vue
+++ b/web/src/components/planner/PlannerFactoryProducts.vue
@@ -125,7 +125,7 @@
         </v-chip>
       </v-row>
       <v-row
-        v-if="Object.keys(product.requirements).length > 0"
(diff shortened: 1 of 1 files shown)


samsmithnz:
@Maelstromeous: not sure how to handle the copilot suggestions above, would value your opinion on how to optimize - or even if optimization is needed, as it's already using `||` - so wouldn't look at the rest of the conditions MOST of the time.

Maelstromeous:
@samsmithnz it's readable enough to me,especially considering it's not offering an alternative...

Maelstromeous (approved):

PR #176, opened 2024-11-25
fix(backend): Increased size of payload bodies

### LINKED ISSUE #172: 413 "Payload too large" when trying to share a late game factory.
Created a late game scenario, 1 of each phase 5 space elevator part, 1 plutonium fuel rod, and 16 factories that make all of the parts from raw materials to the goals listed above. JSON file is about 213KB, large, but not massive. 

[sams-factory.json](https://github.com/user-attachments/files/17896616/sams-factory.json)

Maelstromeous:
Closes #172

### DIFF
--- a/backend/backend.ts
+++ b/backend/backend.ts
@@ -33,8 +33,9 @@ const shareRateLimit = rateLimit({
 });
 
 const app: Express.Application = Express();
-app.use(Express.urlencoded({ extended: true }));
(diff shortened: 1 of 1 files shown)


Maelstromeous:
Tested by loading Sam's big boi factory and increasing the ballastic warp drives, observing locally it was failing pre patch, works post patch.

PR #182, opened 2024-11-25
Add error page when failing to load factory data

### LINKED ISSUE #179: Represent fatal errors better to the user
When the factory plan fails to load, the user is currently presented with a blank page.

In order to fix their data, it generally means they have to blow their local storage away. For the browser novice this is quite a hard thing to do.

Therefore we should present a "Ooops, your plan didn't load. Unfortunately you have to delete your data or restore it from a remote backup". 

Which then blows away the user's localstorage factoryTab and if they're logged in attempt to replace it with the data. If it fails again we need to show them that "Yeah sorry shits corrupted, you need to blow your data away, sorry"

Screenshot 2024-11-25 at 15 12 18

Maelstromeous:
Closes #179 

Intent of this PR is to display an error page to the user prompting them to delete their `localStorage.factoryTab` data in order be able to load the planner again.

Improvement for this would be to allow the user to try and log in, however `Auth` component breaks due to requiring `appStore`, which itself breaks because of the invalid factory data, something custom would have to be created.

For now, the user is prompted to report the error to Discord (with a convient button to copy their data and make it as effortless as possible to report) so we can try to monitor how many breakages we cause.


Maelstromeous:
https://youtu.be/nI-jIB5ycjo

PR #186, opened 2024-11-26
Fixes quantum power calculation

samsmithnz:
The recipe for Neural-Quantum Processor has a minimum power of 0 and a max power of 2000. The Average should be 1000, but it wasn't. Because of this conditional. 
```
if (lowPower && highPower) {
  powerPerBuilding = (lowPower + highPower) / 2;
}
```

Now this is checking to make sure the low and high value is truthy - e.g. not false, 0, "", null, undefined or NaN. See the problem there? I didn't initially... but because the low value is 0 - it was basically considered null, and didn't calculate the average for this recipe. Here is the new code:
```
if (lowPower != null && highPower != null) {
  powerPerBuilding = (lowPower + highPower) / 2;
}
```

### DIFF
--- a/parsing/src/recipes.ts
+++ b/parsing/src/recipes.ts
@@ -119,7 +119,7 @@ function getRecipes(
                 lowPower = Number(recipe.mVariablePowerConsumptionConstant);
                 highPower = Number(recipe.mVariablePowerConsumptionFactor);
                 // calculate the average power
-                if (lowPower && highPower) {
+                if (lowPower != null && highPower != null) {
                     powerPerBuilding = (lowPower + highPower) / 2;
                 }
             }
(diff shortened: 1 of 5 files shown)


Maelstromeous (dismissed):

parsing/src/recipes.ts
```diff
@@ -118,8 +118,8 @@ function getRecipes(
                 // get the power from the recipe instead of the building
                 lowPower = Number(recipe.mVariablePowerConsumptionConstant);
                 highPower = Number(recipe.mVariablePowerConsumptionFactor);
-                // calculate the average power
-                if (lowPower && highPower) {
+                // calculate the average power: Note that because low power can be 0, (and often is), we can't use truthy checks to validate these values
+                if (lowPower != null && highPower != null) {
```
Maelstromeous, line 122:
```suggestion
                if (lowPower !== null && highPower !== null) {
```
No loose typing here please :P

Maelstromeous (approved):
Nice! 👍

PR #188, opened 2024-11-26
Added Imports to show while Factory is hidden

### LINKED ISSUE #6: Imports are not showing up for collapsed / hidden factories
Should be showing imports if applicable, currently only showing exports and products.

AndrewThomasWill:
- Added a section on hidden factories to show 'Import' (modeled just like Exports)

No Imports
image
Factory Satisfied Imports
image
Factory Not Satisfied Imports
image

Closes #6

### DIFF
--- a/web/src/components/planner/PlannerFactory.vue
+++ b/web/src/components/planner/PlannerFactory.vue
@@ -99,7 +99,7 @@
         <!-- Hidden factory collapse -->
         <v-card-text v-show="factory.hidden" class="pa-0">
           <div
-            v-if="factory.inputs.length > 0"
+            v-if="factory.inputs.length > 0 || Object.keys(factory.rawResources).length > 0"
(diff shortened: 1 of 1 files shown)


Maelstromeous:
Nice @AndrewThomasWill ! One extra thing though, would be nice if we can add the Raw Resources to the Inputs. Based on the demo plan if you look at Oil Processing or Copper Basics they have no inputs but they actually do, the raw resources.

Maelstromeous (changes requested):
As above

AndrewThomasWill:
> Nice @AndrewThomasWill ! One extra thing though, would be nice if we can add the Raw Resources to the Inputs. Based on the demo plan if you look at Oil Processing or Copper Basics they have no inputs but they actually do, the raw resources.

image

Pushing the commit now 👍

Maelstromeous (approved):

PR #194, opened 2024-11-26
Improve monorepo CI practices and add pnpm lint(check) commands to all components

### LINKED ISSUE #193: Skip builds when not relevant
At the moment the CI steps are being ran on every single change, including a deployment to vercel when we do a change to the backend for example.

Need to add some filtering to skip things when they're not needed.

Maelstromeous:
Closes #193 

This PR intends to standardize how our component builds are performed, with the objective to introduce pnpm lint to all projects and include it as a mandatory check step.

However, until all the project files have been lint fixed (which will be done as seperate PRs) this is just simply a migration from a unified backend file to a broken down series of files.

Also removed old backend workflows that were non functional. Kept some of them for future reference for automated backend deployments.


samsmithnz:
LGTM :shipit:

Issue #151, opened 2024-11-23
Fix darkreader


Maelstromeous:
The popular plugin Darkreader malforms the site. Need to find what it uses to determine if a site is in dark mode and make it play nice.

![image](https://github.com/user-attachments/assets/9382d7a3-f13e-42df-b640-585d7c2402d8)

Maelstromeous:
This information is very annoyingly hard to find out... it _appears_ to use a reletively new CSS property to figure this out: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

Maelstromeous:
Asked the question here: https://github.com/darkreader/darkreader/discussions/13564

Maelstromeous:
Oh my lord apparently it's added to a list here: https://github.com/darkreader/darkreader/blob/e2557931a23a6a77e651c26c56eb6c8eb2bab09f/src/config/dark-sites.config

Maelstromeous:
Opened PR here with the plugin: https://github.com/darkreader/darkreader/pull/13565

Maelstromeous:
Darkreader PR was merged, unsure when it'll be applied to the plugin though.

Maelstromeous:
This will be fixed next time the plugin gets updated, which should gradually roll out to users. https://github.com/darkreader/darkreader/discussions/13564#discussioncomment-11399624

PR #197, opened 2024-11-27
fix(web): Added sitewide.scss file and applies color-scheme property to all elements

Maelstromeous:
I'm not fully sure this `fixes` #151 but worth a shot...


AndrewThomasWill:
I don't have this extension, so someone else might have to review this 👍

Issue #174, opened 2024-11-25
Performance starts to degrade in larger plans


450Volts:
It is very noticeable after 10 -15 factories. Unusable after 40. Windows 11. Chrome.

# Mael's edits
- We should not be running `calculateFactories` when loading the plan. Loading the plan should literally just be loading the plan.
- When opening a share link, we again should not be running `calculateFactories`.
- When swapping tabs, we should not be running `calculateFactories`
- When editing factories that have no exports or imports, we should not be running calculations upon other factories, as they are not involved (exception to this may be power calculations and world resources)
- We should not be running `calculateFactories` on every single factory, we should only be running `calculateFactories` on factories that are linked to the one that has been updated.
- Deleting a factory makes `calculateFactories` run at least twice. It is running `setFactories` twice for some reason.
- Rearranging the factories calls `calculateFactories`, presumably because setFactories is called.

samsmithnz:
I also noticed some performance lag when I created my 16 factory build. We need to do more analysis at scale - but transparently @450Volts, we have only started working with scale factories in the last couple of days.

450Volts:
> @450Volts can you offer a share link please?

This is the link generated from the share button in mobile. https://satisfactory-factories.app/share/happy-witty-byte

I find that text fields that force updates (Factory name, item numbers, etc.) are the one that lag.

Maelstromeous:
Yeah the thing you're seeing in particular is the fact when you update the name it runs calculations on EVERYTHING. But, rather than making it just do the update when you press enter and call it a day we (us devs) need to look at why it's lagging so hard in the first place, as it also showing for when you swap tabs etc.

Maelstromeous:
@450Volts  Changing the name no longer causes massive lag, we're looking into other areas of improvement.

450Volts:
Confirm! Works much better!

Maelstromeous:
Just got this profile where within a large plan (~40 facs) I was updating the notes of a factory and it was delayed by at least a SECOND PER LETTER!

https://profiler.firefox.com/from-browser/calltree/?globalTrackOrder=201&hiddenGlobalTracks=0&hiddenLocalTracksByPid=7572-0w4&implementation=js&tabID=76&thread=o&v=10

Seems Vue is doing a lot of shit it doesn't need to be doing when doing simple operations as updating the notes view. Need to learn how to make it no propagate like this.

PR #198, opened 2024-11-27
fix(web): Removed updateFactory call when editing names. This will significantly reduce stutter!

Maelstromeous:
#174 Doesn't fix the lag fully but will remove one of the most noticiable elements of it.

Can confirm renaming still works dynamicly (captured on preview site)

https://github.com/user-attachments/assets/aeff1b5f-ba6e-4ea3-96e3-b77bffa8e016


AndrewThomasWill (approved):
🎉

nils-keller-dev (approved):

samsmithnz (approved):

PR #227, opened 2024-11-29
feat(web): Added ability to mark factories as out of sync with the game so the user can check them

### LINKED ISSUE #222: Add "In sync with game" highlighting
A large annoyance of using the tool right now is if you've made changes your plan, either accidentially or intentionally, you need to immediately remember to then apply said changes in game.

If you've forgotten or simply changed too much to keep track, the planner falls out of sync with reality. Which is no fault of the planner, (even if it's a skill issue), but the planner could do a lot better at telling the player "hey now you need to check these factories to ensure they're in sync with reality". The player is in then a constant cycle of updating the plan then having to review their factories to ensure it's correct.

I propose this can be done pretty easily. Initially while the player is planning things, do nothing. But as soon as they check the box of "This factory in the plan is in sync with the game" aka "In sync with game" then whenever the production (and I guess, exports?) of the factory has changed, that checkbox is unticked.

Dead simple, and I hope very useful to people.

In sync factories should probably be in a nice muted green or blue to denote they're in sync and not needed to be looked at.

Maelstromeous:
Closes #222 

This PR enables the user to do the following:

- Mark a factory as "in sync with game" denoting that they have done what the plan says in game
- Mark a factory as "out of sync" as a todo for them to fix later
- When a factory is marked as In Sync, the values of the products are saved. When the products get updated, the factory falls out of sync which the user then needs to put it back in sync.
- This feature is optional, the factory is created as `inSync: null` and when they mark it as sync for the first time does it then start tracking the sync state.
- I've updated the Factory List to be 1) more compact and 2) integrate the sync status.

Video https://youtu.be/dA4O4CMNi1Q

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" style="overflow: hidden">
-
+<html lang="en">
 <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -1,9 +1,5 @@
 <template>
-  <v-app>
-    <v-main>
-      <router-view />
-    </v-main>
-  </v-app>
+  <router-view />
 </template>
 
 <script lang="ts" setup>
--- a/web/src/assets/styles/global.scss
+++ b/web/src/assets/styles/global.scss
@@ -114,7 +114,7 @@
   }
 
   &.needsSync {
-    border: 2px solid #ff8800 !important;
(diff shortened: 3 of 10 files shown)


AndrewThomasWill:
Code Review looks good, Testing on the Preview deployment. I'm hyped to get this to Prod 😄

The only thing I want to keep an eye on is thinking about how to optimize all these calculations in 'calculateFactory' in the future.
Some of them could be combined so we aren't looping through the same dataset multiple times (which at scale could become a lot for many factories). But for now, I like the separation of concerns that we have

AndrewThomasWill (approved):

Issue #245, opened 2024-12-01
Satisfaction: Exports demands are not being represented


Maelstromeous:
When we're looking at the Supply vs Demand in Satisfaction, the demand from Exports is not being shown.

This is confusing. As the user has to look in two places to understand the demands of the factory. Also, when there are export demands, you start seeing numbers like 1000/500 which makes zero sense.

PR #246, opened 2024-12-01
Satisfaction calculations refactor, Satisfaction display tweaks

Maelstromeous:
Large refactor on how Satisfaction is calculated, correcting the inverse mental gymnastics and integrating Export data into the calculation.

It is intended as a prelude for #245, where in the Satisfaction section we are not fully illustrating all supply and demands set upon the factory's parts.

# Visual changes before:
Note how in here Plastic has an export demand but that doesn't show up in Satisfaction.
Screenshot 2024-12-01 at 13 02 21

# Visual changes after:
TBD


samsmithnz:
Suggestion: Consider updating everywhere we show `640/0 /min` to `640/0 per/min`, to break up the large number of `/` that appear when showing parts. I think there is enough space for the extra text and would help with usability - i think!

samsmithnz (approved):

samsmithnz:
Did you try using "Supply and Demand" instead of "S and D"? Was there not enough room?

Maelstromeous:
> Suggestion: Consider updating everywhere we show `640/0 /min` to `640/0 per/min`, to break up the large number of `/` that appear when showing parts. I think there is enough space for the extra text and would help with usability - i think!

Yeah I need to standardize the display of the unit, as sometimes it has a space, sometimes it doesn't.

Maelstromeous:
> Did you try using "Supply and Demand" instead of "S and D"? Was there not enough room?

Yeah, borks quite hard on <1440p unfortunately, not enough width. Will be addressed in the great Satisfaction 3.0 refactor I'm going to do soon.

Issue #46, opened 2024-11-18
Logistic and Storage Center


TripleNipplFish:
Hello,

great work, love the concept. 

Would be cool to have the possibility to make storage and logistic centers where you just can import and export thinks without producing something.

Maelstromeous:
This unfortunately breaks the concept of the site, it's kind of a use case that wasn't really designed for it. The primary reason for this is that storage centers are variable, they contain a lot of products that are ultimately fed by other factories and something has to produce the items that go into it. You loose an element of control of importing rates because the storage bursts it's output due to it's large buffers.

This tool isn't really designed for that concept, unfortunately. I appriciate this isn't the answer you really want, but it would be a substantial code change to support invisible "factories" that contain no products. It basically breaks the accounting math of the site.

In my personal opinion, with the advent of Dimensional Storage, there's no real need to have a centralized storage area any more. You can have up to 5 stacks of any item in the game in the DS, which for me nullifies the need for a physical storage area. It also hides issues with supply and demand by adding a buffer, which is inefficent and adds unneeded complexity when really you should deliver goods directly.

You could always still use your storage area, ultimately the factory producing the items is still delivering somewhere, it's just taking a long time to do so via a proxy.

I'll leave this issue open for a time in case someone has other ideas.

TripleNipplFish:
Thanks for the response.

I’ve reviewed the code a bit. Unfortunately, I’m not a TypeScript or JavaScript Dev at all. For my needs, it seems like imports would need to be included as part of the calculated surplus. However, I assume this would require changes to the factory interface. I’m not asking you to implement this, but if I were to create an implementation where inputs could exceed production usage and be monitored in the export tab, would you consider merging that pull request?

Maelstromeous:
@TripleNipplFish  when you say exceed production usage do you mean you have an import that's purposely higher than the requirements? As that is already possible now, it'll just show Trim next to it.

samsmithnz:
I wonder if part of the answer here to just create a factory called 'storage', that imports in excess production? The only issue is that you need to add the product to the central storage so that you can select it for import, and then remove the product after (unless after adding the product - it first looks to see what is being imported before deciding what dependencies it needs?)

![Image](https://github.com/user-attachments/assets/b644ff10-d5dd-48f4-8e13-f3aca9f5a3e6)
![Image](https://github.com/user-attachments/assets/4837c745-3b34-4958-afd0-ff9f1f4ea650)

[...]

PR #254, opened 2024-12-03
Draft: Adding Sink (Buildings) 

AndrewThomasWill:
- Adding the ability for the user to Add Sink(s) to a Factory

WIP: Parser correctly pulls the building to the gameData, but still need to figure out the remaining UI implementation. We also are likely going in a different direction where the building could create a 'Demand' for a Product

(Initially, that didn't seem valuable since in most scenarios you want to see what is 'available' to export at a glance. Putting some items in a 'Sink' could obscure that information and confuse. However, it's beneficial for the space elevator parts. Those final parts will almost always be going to a Sink. And making sure they have a 'Sink' demand will keep it from showing 'Trim' or 'Not in Demand' for Production)


samsmithnz (dismissed):
It looks good to me - especially from the parser side.

samsmithnz:
I understand this is addressed #7, but I wonder if this should be combined with #46 - it's very close to the same thing - allowing surplus parts to be marked as 'used'.

Maelstromeous:
Clearly sadly this is never going to be completed. Closing. Thank you for the effort long ago @AndrewThomasWill .

PR #280, opened 2024-12-11
fix: moved Fix Production logic into 1 if to ensure there is only one button.

### LINKED ISSUE #240: Products: Fix Production button appears twice when internal and dependencies have insufficent production
https://satisfactory-factories.app/share/absurd-nutritious-answer

One is fixing internal production, the other is fixing exports production. One button should do both.

firefox_Js69Jr6FdG

samsmithnz:
fixes #240

### DIFF
--- a/web/src/components/planner/products/Product.vue
+++ b/web/src/components/planner/products/Product.vue
@@ -61,7 +61,7 @@
           class="rounded mr-2"
           color="green"
           @click="fixProduction(factory, product.id)"
-        ><i class="fas fa-wrench" /><span class="ml-1">Fix Production</span>
+        ><i class="fas fa-wrench" /><span class="ml-1">Fix</span>
         </v-btn>
         <v-btn
           v-show="shouldShowTrim(product, factory)"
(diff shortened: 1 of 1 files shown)


web/src/components/planner/products/Product.vue
```diff
@@ -57,19 +57,12 @@
       </div>
       <div class="input-row d-flex align-center">
         <v-btn
-          v-if="!factory.parts[product.id]?.satisfied"
+          v-if="!factory.parts[product.id]?.satisfied || factory.dependencies.metrics[product.id]?.difference < 0"
           class="rounded mr-2"
           color="green"
           @click="fixProduction(factory, product.id)"
         ><i class="fas fa-wrench" /><span class="ml-1">Fix Production</span>
```
Maelstromeous, line 64: Tbh I'd rename this from Fix Production to simply just "Fix" as it can come from multiple sources

Maelstromeous (changes requested):

Maelstromeous:
Build failing due to redundant import

samsmithnz:
Redundant import build fails: I assume that wasn't caused by this pr, right?

samsmithnz:
Abandoning.

PR #285, opened 2024-12-11
Performance & Dependency Improvements aka "the great lag reducer update"

### LINKED ISSUE #174: Performance starts to degrade in larger plans

Maelstromeous:
Closes #174 

Below examples use my "mega plan" of ~30 factories with lots of interdependencies.

# Pain points

Below are particular pain points of the application where **significant** delays have been noticed for processing.

## `calculateDependencies`
This was causing massive delays as the generation of all the dependencies and their metrics was being re-done every single change regardless of the change and the factories affected.

This process has been changed so that it is ran once upon an update to a factory. Previously, it was also nuking away all of the dependency tree and rebuilding it, now it does not do this and instead checks each input and request pair to ensure they are still valid. For some reason this is a lot faster, probably due to the fact there's less reactivity going on.

## `updateFactory`
This was also being rather suboptimal in how long it took to process a factory, which took around 1s per factory.

Previous updates to `PlannerFactoryInput` made it so that only the relevant factories are called for a recalculation. But, there was much inefficencies between each run.

# Dependency calculation timing refactor
I've moved the dependency calculations to be outside of `calculateFactories` as it mean that said function was recalculating the try upon ALL changes to ANY factory.

`calculateDependencies` is now called on-demand primarily by the `PlannerFactoryInput` component as and when the user does any changes to any factory input, rather than upon every single change. 

This did mean introducing a "load mode" for first time running of the factory plan, template loads and also tests as it created a chicken-and-egg scenario where dependency data depended upon part data being there to test validity, but it wasn't present at runtime, so it has to be run twice, once skipping checks for part data (which as part of the run creates it) and then calculated once more with the part data present to check for invalid links.

# Dependency validity refactor
There were a number of logic errors with the way that the planner was checking if a dependency was invalid. For example it wasn't fully taking into account power producer stuff, it only cared if there was a part (which in of itself is incorrect, just because there's an import to a factory providing a part doesn't mean that it is being exported).


web/src/stores/app-store.ts
```diff
@@ -225,7 +226,7 @@ export const useAppStore = defineStore('app', () => {
     getLastEdit,
     setLastSave,
     setLastEdit,
-    getFactories,
+    getFactories: () => factories.value,
```
Maelstromeous, line 229: @nils-keller-dev Hmm I'd slightly disagree with this one for one primary reason, the current `getFactories()` method ensures that the data is correctly migrated. If we're able to refactor so that when the page loads that the data is fed through the current `getFactories()` method then referencing the value directly thereafter would work. I added the checks in `getFactories` so that it ensures data integrity despite whatever the source may be.

Maelstromeous: Perhaps we create a new one called `initialLoad` or similar

Maelstromeous: Added it in myself

AndrewThomasWill (approved):

Maelstromeous:
Awaiting final testing by @samsmithnz , once he's happy this'll be merged.

web/src/components/planner/PlannerTooManyFactoriesOpen.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div v-if="showAlert" class="count-warning bg-amber w-100 text-center py-1 position-absolute">
+    <span v-if="tooManyShownFactories" class="mr-2"><b>Too many factories are visible (>{{ factoryLimit }})!</b> Please hide some factories or your planner will MAJORLY lag!</span>
+    <span v-if="tooManyInputs" class="mr-2"><b>Too many imports are being shown (>{{ inputLimit }})</b>, which causes large amounts of lag. Please hide some factories!</span>
+    <v-btn density="compact" @click="emit('hide-all')">Hide all</v-btn>
+  </div>
+</template>
```
ClubiNew, line 7: Might be a nice touch to add a close button to this warning for any "power users" that want to keep everything open despite slower performance.

ClubiNew (approved):
LGTM to me, feels a lot snappier now!

samsmithnz (approved):

web/src/components/planner/PlannerTooManyFactoriesOpen.vue
Maelstromeous: Yeah I might put it in later on :)

PR #291, opened 2024-12-15
Allow editing product ingredient amounts

### LINKED ISSUE #173: Products: Allow editable ingredients to scale production to meet ingredient supply
image111111111

I'm trying to maximise the nodes in my vacinity.

In the image (example) have 2 mk4 belts of Iron Ore (960 total), sent to a factory using the Pure Iron Ingot recipe. The suggestion being that the number displayed next to Requires, "Iron Ore: 960.08/min" be editable. This would in my case, save me from guessing what the output would be prior to making the factory or checking after the fact.

I was asked to add an entry on github. Apologies for the terrible wording and hope the suggestion makes sense.

immerrr:
Fixes #173

image

This PR contains a bit of trickery I needed to make component tests work, otherwise they raised errors because:
- game data loading was browser-specific, the easiest way was to refactor it into a module and mock the loader to load from local file during tests instead
- pinia storage was not ready at the moment `@/utils/helpers` was imported, which executed `useGameDataStore()` at the top level, so i added a very early test setup to prepare pinia

I also added names for some of the inputs to simplify selecting them in tests. I don't have a strong opinions on what should be the final names for those.

### DIFF
--- a/web/src/components/planner/products/Product.vue
+++ b/web/src/components/planner/products/Product.vue
@@ -50,6 +50,7 @@
           label="Qty /min"
           :max-width="smAndDown ? undefined : '110px'"
           :min-width="smAndDown ? undefined : '100px'"
+          :name="`${product.id}.amount`"
           type="number"
           variant="outlined"
           @input="updateFactory(factory)"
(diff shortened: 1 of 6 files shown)


web/src/components/planner/products/Product.vue
```diff
@@ -297,4 +338,7 @@
   .product {
     border-left: 5px solid #2196f3 !important
   }
+  span.v-chip.blue div.v-input:deep(input) {
```
Maelstromeous, line 341: Would much prefer a custom class on this one :)

Maelstromeous (changes requested):
Very nice overall! Also I think you just resolved a large headache I had with getting Vue tests to work with components, thank you for this!

Looks like there's some build errors that requires addressing, other than that, looks great!

web/src/components/planner/products/Product.vue
immerrr: Something like this?

Maelstromeous: Was probably better than my way of doing it haha as it's more specific so yes, happy :)

Maelstromeous (changes requested):
Still has build errors unfortunately :( run `pnpm build` locally and it'll tell you what to fix

Maelstromeous:
Just need to run `pnpm lint` to fix the check.

Maelstromeous (approved):
Looks great! Thank you @immerrr!

PR #296, opened 2024-12-16
fix: bug with unit when there is a large negative power

samsmithnz:
This pull request includes a change to the `formatPower` function in the `web/src/utils/numberFormatter.ts` file. The change updates the logic for converting units to gigawatts.


Maelstromeous (approved):
Good spot :)

PR #297, opened 2024-12-16
Imports: Enable selection of the same part across multiple factories

### LINKED ISSUE #290: Can't import the same (by)product from two different factories
Hello yall,
with version 2 i could simulate my blueprints by copying small factories and importing one after the other in the factory that uses their product. This doesn't seem to be possible any more with v3. First I thought, my factory data are corrupted. So I recreated everything but still could reproduce the problem, see screenshot.

grafik

I propose, that users can decide to import the same product from different factories. But maybe it is a bug.

The console shows the following error, although the products and byproducts of the factory 8611 are correctly stated in the json (enclosed):

```
calculatePartSupply - inputs: Output part is missing from input! 
Proxy { <target>: {…}, <handler>: {…} }
​<target>: Object { factoryId: 8611, outputPart: null, amount: 0 }
​​amount: 0
​​factoryId: 8611
​​outputPart: null
​​
```
[2024-12-15-satisfactory-factories.json](https://github.com/user-attachments/files/18140643/2024-12-15-satisfactory-factories.json)

I wanted to import heavy oil residue from all the plastic factories.

Thanks for your time, 
best regards
Zekonnd

Maelstromeous:
Closes #290

`importPartSelections` now determines selected parts via factory-part combinations, preventing same-selection of parts on the same factory rather than unique across all factories by part.

Below used to show up `IronIngot` and also without a selector on the second factory, now shows up correctly:
Screenshot 2024-12-16 at 22 30 21


Maelstromeous:
Pushing this one through, works very well and it's quite a critical one.

Issue #299, opened 2024-12-16
META: Trim, Satisfy and Fix Production logic corrections


Maelstromeous:
Various buttons have drifted from their original purpose and are now in a currently buggy state. Since there are multiple issues to this effect, this issue will be used to track them all and other edge cases we've discovered.

## Inputs
- #324
- #242 
- #251 

## Products
- #271
- #321 
- #240 

## Satisfaction
- #220

Maelstromeous:
Happy this is now corrected. Closing.

PR #301, opened 2024-12-16
fix(web): fix double "fix production" button

### LINKED ISSUE #240: Products: Fix Production button appears twice when internal and dependencies have insufficent production

nils-keller-dev:
Closes #240


web/src/components/planner/products/Product.vue
```diff
@@ -57,17 +57,10 @@
...
-          @click="fixProduction(factory, product.id)"
-        ><i class="fas fa-wrench" /><span class="ml-1">Fix Production</span>
-        </v-btn>
-        <v-btn
-          v-if="factory.dependencies.metrics[product.id]?.difference < 0"
-          class="rounded mr-2"
-          color="green"
-          @click="fixExport(factory, product.id)"
+          @click="onClickFixProduction(factory, product.id)"
         ><i class="fas fa-wrench" /><span class="ml-1">Fix Production</span>
```
Maelstromeous, line 64:
```suggestion
        ><i class="fas fa-wrench" /><span class="ml-1">Fix</span>
```

Maelstromeous:
Looks like there's a bug @nils-keller-dev:

Screenshot 2024-12-17 at 00 00 59
Screenshot 2024-12-17 at 00 01 02

Repro:

1. Load Demo template
2. Look at Copper Basics
3. Note how wire is satisfied but it's being asked to be fixed, pressing on fix does nothing.

Also I think we can do this simplier. `factory.parts` all have parts with a `satisfied` key. It uses both internal product demands and export demands to create that value, so we should be able to just check if the part is satisfied as the condition of showing it.

Additionally, in the parts array, `amountRequired` is encompassing both internal production and external demands. It should be literally a matter of just making production match the `amountRequired` value.

Maelstromeous (changes requested):
See above

nils-keller-dev:
> Also I think we can do this simplier. `factory.parts` all have parts with a `satisfied` key. It uses both internal product demands and export demands to create that value, so we should be able to just check if the part is satisfied as the condition of showing it.

So we don't need to check `factory.dependencies.metrics[id]?.difference`?


> Additionally, in the parts data, `amountRequired` is encompassing both internal production and external demands. It should be literally a matter of just making production match the `amountRequired` value.

That's what is happening in `fixProduction()`, right? So I don't need the extra logic regarding `fixExport()`?

Maelstromeous:
> So we don't need to check `factory.dependencies.metrics[id]?.difference`?

Don't think it's nessessary, no, it should be possible purely off amountRequired as that's a calculated value derived from the other demand sources (production, exports)

> That's what is happening in `fixProduction()`, right? So I don't need the extra logic regarding `fixExport()`?

I believe that's the case yes, requires testing naturally.

Maelstromeous:
Related #299

PR #302, opened 2024-12-17
Allowing Factories with only Power Gens to Import

AndrewThomasWill:
- Added a case for allowing Imports if the Factory only has a Power Gen
- Added a Template to let us load this example in 'Debug'
- Added a unit test for the 'Inputs' to cover this scenario


Maelstromeous:
Noting here that "Add Import" button is showing up when the selections have been exhausted. It's kiiiinda out of scope of this PR but we may have to create another issue to formally address it. Probably for one when #285 is merged.

As now, this occurs:

Screenshot 2024-12-17 at 00 15 04

Maelstromeous (changes requested):
Minor tweaks otherwise good :)

AndrewThomasWill:
> Noting here that "Add Import" button is showing up when the selections have been exhausted. It's kiiiinda out of scope of this PR but we may have to create another issue to formally address it. Probably for one when #285 is merged.
> 
> As now, this occurs:
> 
> Screenshot 2024-12-17 at 00 15 04
> #303

image

Should be fixed in the latest commit. Also cleaned up some of the logic to reuse the 'No Products' type error message

web/src/components/planner/PlannerFactoryImports.vue
```diff
@@ -140,7 +140,7 @@
           @click="addEmptyInput(factory)"
         >Add Import
         </v-btn>
-        <span v-if="ableToImport(factory) === 'noProducts'" class="ml-2">(This factory does not yet have any products.)</span>
+        <span v-if="ableToImport(factory) === 'noProductsOrProducers'" class="ml-2">(This factory does not yet have any products/power producers.)</span>
```
Maelstromeous, line 143:
```suggestion
        <span v-if="ableToImport(factory) === 'noProductsOrProducers'" class="ml-2">(This factory does not yet have any products/power generators.)</span>
```

web/src/components/planner/PlannerFactoryImports.vue
```diff
@@ -140,7 +140,7 @@
           @click="addEmptyInput(factory)"
         >Add Import
         </v-btn>
-        <span v-if="ableToImport(factory) === 'noProducts'" class="ml-2">(This factory does not yet have any products.)</span>
+        <span v-if="ableToImport(factory) === 'noProductsOrProducers'" class="ml-2">(This factory does not yet have any products/power generators.)</span>
```
Maelstromeous, line 143: Yeah I think you're right on this one, I don't think this ever becomes true.
```suggestion
```

Maelstromeous (dismissed):

Maelstromeous (approved):

PR #313, opened 2024-12-24
Parsing data corrections

samsmithnz:
Various data corrections to match data in original raw data file and [official wiki](https://satisfactory.wiki.gg/wiki/Crude_Oil),  correct sorting, and refactored a few variables to make their purpose clear. 

These were discovered as a giant cross check/ deep dive into the data, and is a stepping stone before additional refactoring - as this ensures we start with the correct data.

Follow up is that 2 part images needed to be renamed (it's slightly weird we use the display name, and not the part id). Instead of deleting the old part images, I've kept them, as I believe this will help with backward compatibility until the users cache refreshes.


parsing/src/parts.ts
```diff
@@ -220,7 +229,7 @@ function getItems(data: any[]): ParserItemDataInterface {
                     energyValue *= 1000;
                 }
 
-                //console.log(`Adding part: ${partName} (${friendlyName}) with energy value: ${energyValue}`);
+                // console.log(`Adding part: ${partName} (${friendlyName}) with energy value: ${energyValue}`);
```
samsmithnz, line 232:
```suggestion
                //console.log(`Adding part: ${partName} (${friendlyName}) with energy value: ${energyValue}`);
```

parsing/src/buildings.ts
```diff
@@ -47,10 +47,11 @@ function getPowerConsumptionForBuildings(data: any[], producingBuildings: string
         .filter((entry: any) => entry.Classes)
         .flatMap((entry: any) => entry.Classes)
         .forEach((building: any) => {
-            if (building.ClassName && building.mPowerConsumption) {
+            if (building.ClassName && building.mPowerConsumption >= 0) {
```
samsmithnz, line 50:
```suggestion
            if (building.ClassName && building.mPowerConsumption) {
```

parsing/gameData.json
```diff
@@ -501,11 +502,11 @@
                 "energyGeneratedInMJ": 750
             },
             "LiquidOil": {
-                "name": "Liquid Oil",
+                "name": "Crude Oil",
                 "stackSize": 0,
                 "isFluid": true,
                 "isFicsmas": false,
-                "energyGeneratedInMJ": 0
+                "energyGeneratedInMJ": 320
```
Maelstromeous, line 509: We sure this actually generates any energy? Can it be burned directly?

parsing/src/recipes.ts
```diff
@@ -218,15 +218,15 @@ function getPowerGeneratingRecipes(
                 const primaryFuelPart = parts.parts[primaryFuel];
 
                 const burnDurationInMins = primaryFuelPart.energyGeneratedInMJ / burnRateMJ;
-                const burnDuration = burnDurationInMins * 60; // Convert to seconds
+                const burnDurationInS = burnDurationInMins * 60; // Convert to seconds
```
Maelstromeous, line 221: Is this not meant to be a divide?

Maelstromeous (commented):
Left some queries

parsing/src/recipes.ts
samsmithnz: This is your code. Lol, I was just adding the unit to the variable. Multiplication is right here!

parsing/gameData.json
samsmithnz: You are correct - today it can not be burned directly - however, it has this value in the game data. Given this, it is possible it could happen in the future. The goal here was to match the game data files - who knows what is coming in 1.1.

samsmithnz: wiki entry for a visual confirmation: https://satisfactory.wiki.gg/wiki/Crude_Oil

samsmithnz: reference in original data: 

image

parsing/src/recipes.ts
Maelstromeous: I think what's confusing me is the variable names.

`burnDurationInMins` looks like it should be `burnGenerationInS` and then what is currently `burnDurationInS` be simply `burnGenerationPerMin`.

Maelstromeous:
Refencing PR description: 

> (it's slightly weird we use the display name, and not the part id)

Unless we batch rename every single image to use part ID, it'll remain as the original naming scheme intended. It may be better to rename them eventually as it'll mean all cache issues are bust. For now, I've removed the year long TTL from the cache backend and reduced it to 1 week.

Maelstromeous (approved):

PR #369, opened 2025-01-25
World Data Import

vassbo:
> Import a .sav world file to see where all the somersloops are.

### What's added
- An "Import World" button
image
- Info popup
image
- When selecting the .sav file [a custom package](https://github.com/vassbo/sav2json) I made will convert the data into a more readable JSON format
- Loading progress _(I made this universal based on your existing startup loading so it can be used for any future loading messages)_
image
- Data popup _(Will show machine type and its product, and also how many somersloops/power shards in the inventory)_
image

- Auto check and update the factories synced state based on what is actually built in game.
The user will have to confirm what building are in one single factory as it can be hard to determine automatically, as most buildings are connected. I have two ideas for that: a world map with the machines placed based on it's coordinates, or probably easier is to show a node graph of all the buildings similar to the current factories graph where the user can select all of the machines in each factory.
_Some of the code for this is already started_ 
- Create a factory plan from scratch from an existing world


Maelstromeous:
Oh wow, this is awesome!

Thank you so much for this contribution, I'll give it a proper test and review possibly tomorrow, I'm currently in the process of announcing Alpha 4 to the world (creating video, change log etc) so will give this my full attention after!

Maelstromeous:
Hey @vassbo, I took a quick look at this and unfortunately couldn't get it to work :(

When I upload this save file: https://www.dropbox.com/scl/fi/p58p4xh9em5xiahrgoi52/A-new-beginning-1.0-_autosave_0.sav?rlkey=e1xq5oxvxbqvr4y5f08viaw3o&st=sr9um7n5&dl=0 (couldn't upload it directly via GH sorry) I get the below error:

image

Hope this helps :)

Otherwise I like the concept and idea of this, and I'm wondering how much we can extrapolate out of the game's data and apply to the planner. Would be amazing for if example we can figure out groups of factories geographically and paste them into the planner as factory groups. Would something like that be possible?

vassbo:
Okay, that looks fixable. I'll try to fix it later today!

We do have access to all of the data so we know the position of all the machines, meaning we can create a system to find factories based on a proximity to nearby machines. This won't work if multiple factories are built very close to each other or on top of each other though.

web/src/components.d.ts
```diff
@@ -18,6 +18,7 @@ declare module 'vue' {
     Introduction: typeof import('./components/planner/Introduction.vue')['default']
     JoinDiscord: typeof import('./components/JoinDiscord.vue')['default']
     KoFi: typeof import('./components/ko-fi.vue')['default']
+    Loader: typeof import('./components/Loader.vue')['default']
```
Maelstromeous, line 21: Very minor thing but could we call this `SaveLoader`?

vassbo:
Got it fixed! :)

Maelstromeous:
Great, works for me now too!

A couple of adjustments I would like to ask for if I may:

1. Could you adjust the "Import World" button to say "Import World [WIP]" as to be fair it is a WIP feature.
2. In the first popup, could you add explanation to better describe that this is a WIP feature, and that it will be expanded to include extra features given time, e.g. importing of factory groups"?

vassbo:
Done!

Maelstromeous (approved):

PR #413, opened 2025-06-29
added reset button to reset sync status

### LINKED ISSUE #397: GameSync: Add ability to reset sync state
Currently you have to clone an entire factory to reset the sync state. Should add a X or a refresh button next to the sync state buttons to reset it.

Resetting it is simply just emptying both `syncState` keys from the factory.

samsmithnz:
fixes #397 

This pull request enhances the functionality of the `PlannerFactory.vue` component by introducing a new feature to reset the sync status of a factory. It includes UI updates and the addition of a helper method to handle the reset action.

### DIFF
--- a/web/src/components/planner/PlannerFactory.vue
+++ b/web/src/components/planner/PlannerFactory.vue
@@ -33,12 +33,30 @@
                 <v-chip class="sf-chip small green no-margin" @click="setSyncState(factory)">
                   <i class="fas fa-check-square" />
                   <span class="ml-2">In sync with game</span>
+                  <v-btn
+                    icon
+                    size="x-small"
+                    class="ml-2"
+                    title="Reset sync status"
+                    @click.stop="resetSyncState(factory)"
+                  >
+                    <i class="fas fa-times" />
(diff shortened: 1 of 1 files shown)


web/src/components/planner/PlannerFactorySatisfaction.vue
```diff
@@ -30,12 +30,20 @@
         :model-value="showSatisfactionBreakdowns"
         @change="changeSatisfactionBreakdowns"
       />
+      <v-switch
+        v-model="hideInternalOutputs"
+        class="ml-4"
+        color="primary"
+        hide-details
+        label="Hide Internal Products"
```
Maelstromeous, line 38:
```suggestion
        label="Hide Internal Products & Raw"
```
Your code currently also hides Raw, which in this context does make sense, so just adjusting the label to match the behaviour.

Maelstromeous (changes requested):
Great addition! One minor thing to change then it can be merged. Eyeball tested it as well.

Maelstromeous (approved):

Issue #398, opened 2025-02-12
Deleting a factory displays multiple error messages


samsmithnz:
Repro
1. open the factories app in a private window
2. select the demo plan
3. delete the circuit board factory
4. See the error messages for oil processing and computer factories being corrupt

Maelstromeous:
Closed via https://github.com/satisfactory-factories/application/commit/2d8bbd591a33b8484c5b292668ccdbd0ef2edb76

PR #435, opened 2026-07-19
fix(web): #398 Deleting a factory no longer displays false corruption error messages

Maelstromeous:
When a factory was deleted, removeFactoryDependants only cleaned up one direction of the dependency graph: inputs on factories that imported FROM the deleted factory. Factories the deleted factory imported from (its providers) still held dependencies.requests entries keyed by the deleted factory's ID. The subsequent recalculation ran flushInvalidRequests, which found requests pointing at a now-nonexistent factory and alerted "corrupted data" once per affected provider factory.

removeFactoryDependants now also removes the deleted factory's requests from all remaining factories, so a normal deletion cannot trip the corruption alert (which remains in place for genuinely corrupt saves). Also fixes an adjacent bug where the dependant loop deleted requests[factory.id] instead of requests[dependentId].


Maelstromeous:
Fix confirmed via preview.

PR #439, opened 2026-07-27
feat(backend): Automate deployment on merge to main

### LINKED ISSUE #438: Automate backend deployment (and fix the three things that stop it working)
## Problem

Shipping the API is a manual `backend/publish.sh` run from a laptop. `docs/how-do-we-release.md` already flags this as not ideal — it means one person can deploy the backend, and the frontend's trunk-based flow stops at the API boundary.

With significant sync work coming, this needs to be automated first, not after.

## What's actually broken

While scoping this, three problems turned up that mean the current deploy path could not work even manually:

1. **The backend image cannot be built.** `backend/package.json` moved to pnpm `catalog:` versions, which only resolve when pnpm can see `pnpm-workspace.yaml`. `publish.sh` builds from `backend/`, so `pnpm install` fails with `ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC`. Reproduced locally. The image in production predates the catalog, which is why nobody has hit it.
2. **The server's compose port mapping is wrong.** 618e944 moved the API from 3001 to 3010, but the compose file still publishes `3001:3001`. Production only works because it's running an image from before that commit — the first rebuild from `main` would publish a container nothing upstream can reach.
3. **A deploy cannot report failure.** The webhook returns `200` before the SSH happens, and `update.sh` on the box exits `0` as long as `docker compose up` returned, even if the container immediately died.

## Proposal

- `Backend: Deploy` on merge to `main`: checks → build & publish image → trigger webhook.
- Publish to Docker Hub `maelstromeous/satisfactory-factories` (matching the treatment given to the other projects on this webhook server), with a `backend-<sha>` tag alongside `backend-latest` for rollback.
- Validate the Docker build on PRs so it can never silently rot again.
- Healthcheck-gated deploy on the box, and docs covering the chain, the secrets and how to verify a deploy actually landed.

Maelstromeous:
Closes #438.

Shipping the API was `backend/publish.sh` run from a laptop — the one thing in this project only one person could do. With a lot of sync work coming, this gets the deployment story sorted first.

Merging to `main` now does the whole thing:

```
main (paths: backend/**, pnpm-workspace.yaml, package.json, .dockerignore)
  └─ Backend: Deploy
     ├─ Checks         lint-check + tsc
     ├─ Publish image  → maelstromeous/satisfactory-factories:backend-latest
     │                                                       :backend-<sha>
     └─ Deploy         → POST <WEBHOOK_URL>
                          └─ webhooks box → ssh sf → /root/update.sh
```

The companion PR that registers the hook is [Maelstromeous/webhooks#9](https://github.com/Maelstromeous/webhooks/pull/9) — **that one has to be merged and rolled out first**, or this pipeline's last step 404s.

---

## ⚠️ Three things were already broken

None of this works without fixing them, so they're in here rather than in follow-ups.

### 1. The image could not be built at all

`backend/package.json` pins `ts-node`, `typescript` and the eslint packages with `catalog:`. A catalog only resolves when pnpm can see `pnpm-workspace.yaml`, and `publish.sh` built from `backend/`. Reproduced on `main` before touching anything:

```
Step 5/7 : RUN pnpm install
[ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC] No catalog entry 'eslint' was found for catalog 'default'.
The command '/bin/sh -c pnpm install' returned a non-zero code: 1
```

The image running in production is 19 months old and predates the catalog, which is why this has sat there unnoticed. `backend/Dockerfile` now builds from the workspace root (`-f backend/Dockerfile .`), with a new root `.dockerignore` so the context isn't the entire monorepo.

Also switched to `node:24-slim` + corepack (pins pnpm to the root `packageManager` field) and `--frozen-lockfile`, so a lockfile that drifts from `package.json` fails the build loudly instead of resolving something untested.

Dev dependencies are still installed deliberately — `pnpm start` is `ts-node backend.ts`, so `ts-node` and `typescript` are runtime dependencies in all but name. There's a comment saying so, because `--prod` is the obvious "optimisation" that would break production.

### 2. The port disagreed with itself

`618e944` ("Changed backend API port to 3010") moved the app but moved nothing else with it: the server compose still published `3001:3001`, and `web/src/config/config.ts` still pointed local dev at `3001`. Production only works because it runs an image from before that commit — `docker logs sf-backend | head` on the box says `Webserver running at http://localhost:3001/` while `main` said `3010`.

**The first successful rebuild from `main` would have published a container nothing upstream could reach — and the deploy would still have gone green.**

Fixed by putting the app **back on 3001** rather than mapping between two numbers. It is now 3001 in `backend.ts`, `EXPOSE`, both compose files, the host port and the Cloudflare tunnel origin, with no translation anywhere. Bonus: the box's existing `3001:3001` line needs no change, so that's one less manual step that could be missed.

> **Known overlap, decided.** `web/testing/global-setup.ts` also binds 3001 for its gameData fixture, and it *silently skips startup* if the port is taken — so running the backend during a `web` test run makes the suite fetch game data from the API, 404, and fail confusingly. That's plausibly why 618e944 moved the API off 3001, though it left the job half-done.
>
> **Call made: 3000 is the web app, 3001 is the API, and those two are fixed.** The overlap is rare enough to live with — we aren't running tests concurrently with a dev backend in practice. If it ever does start to bite, the *fixture* moves to a port of its own (~3005); the API does not move. `PORT` gives a local escape hatch in the meantime (`PORT=3011 pnpm dev:backend`), and the README records the allocation.

### 3. A failed deploy could not report itself

Expanded considerably — see the new **[What was wrong with the old `update.sh`](docs/deployment.md)** section, which walks through all seven failure modes. Summary of what the old three-command script did wrong:

| # | Problem |
| --- | --- |
| 1 | `down && up` destroyed the container on **every** call — ~10s of 502s per deploy (ts-node compiles on boot), including the many no-ops where the pull changed nothing |
| 2 | Couldn't tell whether the container *survived*. `up -d` returns on **started**, not listening; with `restart: always`, a crashlooping container is indistinguishable from a healthy one |
| 3 | Printed `Container updated!` unconditionally — the single line that hid the registry drift for months |
| 4 | Left **no trace** on failure. `set -e` just ended the script, so the log's last line was `Updating container...`, reading like a truncated success. And the failure can land *between* `down` and `up` — API off, indefinitely, log silent |
| 5 | Timestamps went to `deploy.log`, but the compose commands' error text went to the webhook daemon's log **on a different box** |
| 6 | Nothing serialised deploys — the GitHub concurrency group can't stop a workflow colliding with a manual `publish.sh` |
| 7 | Orphaned images accumulated forever (the box has images going back 21 months) |

The rewrite recreates only on a digest change, gates on the healthcheck via `up -d --wait`, logs before/after image IDs, traps `ERR` with exit code and line number, `tee`s everything into one log, takes an `flock`, prunes dangling images older than a week, and reports the container's health state — which surfaces the case where the box's compose has no healthcheck and `--wait` quietly degrades to "is it running".

This script is the **only** component in the chain that can detect a failed deploy: GitHub reports on the POST, the webhook returns 200 before the SSH, and the webhooks server doesn't surface the remote exit code. Whatever it fails to write down isn't recorded anywhere.

---

## Registry: GHCR → Docker Hub

`maelstromeous/satisfactory-factories:backend-latest`, matching how the other projects on this webhook server are set up. A `backend-<sha>` tag is published alongside so a bad deploy rolls back to an exact commit without a rebuild (documented in `docs/deployment.md`).

The `backend-` prefix is kept from the old GHCR tag, so the repo stays free for other components later.

## Also in here

- `.github/workflows/WIP/` — four inert GHCR stubs (one of them empty) superseded by this.
- `build-backend.yml` loses its `push: main` trigger — on `main` it's now called by `Backend: Deploy`, so keeping it would run the same checks twice on the same commit.
- `pr-docker-backend.yml` builds the image on PRs without pushing, so this can't silently rot again.
- `docker-compose-packaged.yml` follows the Dockerfile to the root context (and its `.:/app` volume mount would have shadowed the app, which now lives at `/app/backend`).
- `docs/deployment.md` — the full chain, secrets, rollback, the `update.sh` post-mortem, and a symptom→cause table. `how-do-we-release.md`, `README.md`, `CLAUDE.md` and `architecture/overview.md` no longer say the backend is deployed by hand.

No `CHANGELOG.md` entry: it mirrors the user-facing in-app changelog and none of this is player-visible.

## Validation done

Image:
- ✅ Reproduced the `catalog:` build failure on `main`; new Dockerfile builds clean from the repo root.
- ✅ Built image boots and serves on **3001**: `/hello` → `{"message":"Hello, the server is running!"}`. `ts-node` copes with TypeScript 6.0.3.
- ✅ The compose healthcheck turns the container `healthy` in ~7s.
- ✅ `backend` lint-check (0 errors) and `tsc` pass after the port change.

`update.sh` — exercised against a throwaway local registry, all seven scenarios:
- ✅ **fresh deploy** → `Image updated: (none) -> 9538896912ac`, `Container state: healthy`
- ✅ **no-op** → `Image unchanged`, ~1s, container never recreated
- ✅ **real update** → `Image updated: 9538896912ac -> 811fed4df13a`
- ✅ **failed pull** → `DEPLOY FAILED (exit 1) at line 68.`
- ✅ **container dies on boot** → `container is unhealthy` + `DEPLOY FAILED (exit 1) at line 82.` (the old script exited 0 here)
- ✅ **compose without a healthcheck** → `Container state: NO HEALTHCHECK — --wait only checked the container is running`
- ✅ **concurrent deploys** → second run waited the full timeout then `Another deploy has held the lock`

**Not validated:** nothing has been pushed to Docker Hub and no deploy has run against the real box. The chain is unproven end to end until it's exercised.

## Before merging — manual steps

1. **Merge and roll out [webhooks#9](https://github.com/Maelstromeous/webhooks/pull/9)** — `ssh webhooks && cd /root/webhooks && ./sync.sh`, and check `satisfactory-factories` is in the hook list it prints.
2. **Repository secrets** — this repo currently has **none** set (`gh secret list` returns empty; albion-mapper has all four):
   | Secret | Value |
   | --- | --- |
   | `DOCKERHUB_USERNAME` | `maelstromeous` |
   | `DOCKERHUB_TOKEN` | Docker Hub token with write access to `maelstromeous/satisfactory-factories` |
   | `WEBHOOK_URL` | The deploy hook endpoint (kept out of this public repo — see the private webhooks repo) |
   | `WEBHOOK_SECRET` | the shared webhook secret |
3. **Update the box** — neither file is synced by any deploy:
   ```bash
   scp backend/update.sh sf:/root/update.sh && ssh sf 'chmod 755 /root/update.sh'
   # then in /root/docker/docker-compose.yml:
   #   image: maelstromeous/satisfactory-factories:backend-latest   ← was ghcr.io/...
   #   + the healthcheck block from backend/docker-compose-server.yml
   #   ports: "3001:3001" is already correct — leave it
   ```
   If the image line isn't changed, the hook fires, `update.sh` logs `Image unchanged`, and the API keeps running the 19-month-old GHCR image.
4. **Then `workflow_dispatch` "Backend: Deploy"** and confirm on the box, not in Actions:
   ```bash
   ssh sf 'tail -20 /root/deploy.log'   # want "Image updated: … -> …" and "Container state: healthy"
   curl -s https://api.satisfactory-factories.app/hello
   ```

## Noticed, deliberately not fixed here

- Moving web's vitest gameData fixture off port 3001 (see the note in §2) — a `web` change.
- The image ships `ts-node` and the full dev dependency tree. Compiling to `dist/` and running `node dist/backend.js` would be leaner and faster to start, but it's a runtime change with real risk and it isn't what this PR is for.


Maelstromeous:
### Status check before review

**CI is green**, including the new PR-side Docker build — which is the meaningful one, since it proves the `catalog:` fix works in Actions and not just on my machine:

| Check | Result |
| --- | --- |
| Build Backend | pass |
| build / Build API image | pass |
| Vercel | pass |
| Vercel Preview Comments | pass |

**Docker Hub:** `maelstromeous/satisfactory-factories` already exists and is public, with **0 tags** — so the first deploy populates an empty repo rather than creating one. Nothing to set up there.

**Repository secrets: all four are missing.** `gh secret list` on this repo returns nothing; `dignityofwar/albion-mapper` (the working reference) has all four. So these need adding before the first deploy, or the publish step fails on Docker Hub login and the deploy step posts to an empty URL:

- `DOCKERHUB_USERNAME`
- `DOCKERHUB_TOKEN`
- `WEBHOOK_URL`
- `WEBHOOK_SECRET`

Full ordering is in the "Before merging — manual steps" section of the PR description.

PR #466, opened 2026-07-28
chore(repo): update eslint monorepo to v10

Maelstromeous:
Supersedes #463, which Renovate could not carry on its own: the major needs a config refactor, and its lockfile update was incomplete.

## Why #463 is red

Renovate bumped the catalog (`eslint: 9.39.5 → 10.7.0`) and `@eslint/js` in backend/parsing, but left `web/pnpm-lock.yaml` untouched. The frozen install then fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` — in **all three** jobs, since each one installs the whole workspace. All three lockfiles are regenerated here.

## The actual v10 blocker

Backend and parsing needed no config changes at all. Web did:

`neostandard` (the style layer, standard's flat-config successor) is still ESLint 9-only at 0.13.0 and bundles `@stylistic/eslint-plugin` v2, which calls `sourceCode.isSpaceBetweenTokens` — one of the deprecated `SourceCode` methods ESLint 10 removed. Every lint run crashed with `TypeError: sourceCode.isSpaceBetweenTokens is not a function`.

So web now depends on `@stylistic` v5 directly (the first line that supports ESLint 10) and `eslint.config.mjs` hands that instance to neostandard's configs in place of the bundled v2. v5 dropped three of the rule configs neostandard emits, which the same pass patches — keeping the rules rather than losing them:

| neostandard emits | patched to | why |
| --- | --- | --- |
| `@stylistic/func-call-spacing` | `@stylistic/function-call-spacing` | renamed in v3 |
| `object-property-newline`'s `allowMultiplePropertiesPerLine` | `allowAllPropertiesOnSameLine` | deprecated alias, dropped from v5's schema |
| `@stylistic/jsx-indent`, `@stylistic/jsx-props-no-multi-spaces` | removed | deprecated aliases of `indent` / `no-multi-spaces`, both already configured; this project has no JSX |

I checked all 61 stylistic rules neostandard configures against v5's schemas — those are the only three that don't validate. Web's own `@stylistic/quotes` also moves from `allowTemplateLiterals: true` to `'always'`, the boolean form being deprecated in v5. The whole shim comes out once neostandard supports ESLint 10.

### Why not a pnpm override

[...]

PR #471, opened 2026-07-28
chore(web): pin pnpm for the Vercel build

Maelstromeous:
Follow-up to #466, where this bit us for real.

## The problem

Vercel builds with `web` as the project's root directory, so the `package.json` it consults for corepack is `web/package.json` — not the repo root's, which is where `packageManager` lives. Every preview build therefore logs:

```
WARNING! Could not enable corepack because package.json is missing "packageManager" property
Detected `pnpm-lock.yaml` 9 which may be generated by pnpm@9.x or pnpm@10.x
Using pnpm@9.x based on project creation date
```

So the deployed site is installed with **pnpm 9**, while CI, the API image and any laptop use the **11.11.0** the root pins. Vercel is picking its version from the lockfile format and the project's creation date — not from anything in this repo.

That gap isn't theoretical. In #466 I added an `overrides` block to `pnpm-workspace.yaml`; overrides in that file are pnpm 10+ syntax, so pnpm 9 didn't see them, found one recorded in the lockfile, and killed the deployment with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` — while every other environment installed happily. Any future workspace setting introduced after pnpm 9 can do the same thing, and it will always show up as a broken preview rather than a failing test.

## The change

One line: `"packageManager": "pnpm@11.11.0"` in `web/package.json`, the same version the root declares, so there's nothing to drift between them.

## Verification

- `pnpm install --frozen-lockfile` in `web` still installs clean and leaves `web/pnpm-lock.yaml` untouched — the field is inert for local and CI installs, which already resolve pnpm from the root
- `pnpm build` (`vue-tsc --noEmit` + vite) and `pnpm lint-check` both pass

Whether Vercel actually picks it up depends on `ENABLE_EXPERIMENTAL_COREPACK=1` being set for the project — the wording of the warning suggests corepack is already being attempted and only the field is missing, but the preview build on this PR is the real test. Worth a glance at its log for the `Using pnpm@...` line before merging; if corepack still doesn't engage, the env var needs setting in the project settings and I can follow up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V11Y7TknyJmt75S5enzFWB)_


Maelstromeous:
Marking this draft — the pin works, but it surfaces a second problem that needs a Vercel project setting before this can merge.

Corepack now engages properly:

```
> Detected ENABLE_EXPERIMENTAL_COREPACK=1 and "pnpm@11.11.0" in package.json
! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-11.11.0.tgz
```

and then pnpm 11 refuses the install:

```
[ERR_PNPM_LOCKFILE_CONFIG_MISMATCH] The current "catalogs" configuration doesn't match the value found in the lockfile
```

**Why:** the build only has `web/` — `pnpm-workspace.yaml` isn't in it. I reproduced it exactly by copying `web/package.json` + `web/pnpm-lock.yaml` into an empty directory with no workspace file and running `pnpm install --frozen-lockfile`; same error, every time. With the workspace file present it installs clean.

pnpm 9 got away with this because it read the catalog snapshot baked into `web/pnpm-lock.yaml` (`catalogs.default`: eslint, ts-node, typescript) and never compared it against workspace config. pnpm 11 does compare — finds no catalog config at all, sees three entries in the lockfile, and stops. It's the same mechanism that broke #466's `overrides` block, just a different key: **anything declared in `pnpm-workspace.yaml` is invisible to the Vercel build.**

That makes this a two-part change, and only one part lives in the repo:

1. Enable **Settings → Build & Deployment → "Include source files outside of the Root Directory in the Build Step"** for the project, so `pnpm-workspace.yaml` reaches the builder.
2. Merge this PR, so the build uses the pinned pnpm 11 instead of pnpm 9.

Order matters — this PR alone breaks previews, the setting alone is harmless. Ready to go once the setting is flipped; happy to re-run and confirm the build is green then. If you'd rather not touch project settings, closing this is a perfectly good answer too: pnpm 9 keeps working, at the cost of Vercel staying on a different pnpm from everything else and workspace-level config continuing to be a trap.

For the record, #466 is unaffected either way — its `web/pnpm-lock.yaml` catalog snapshot only changes the eslint version already in there, and its preview deployed successfully.

---
_Generated by [Claude Code](https://claude.ai/code)_

Maelstromeous:
Setting is flipped and this is green — the deployment now gets through install and build instead of dying in `pnpm install` within seconds, so the `catalogs` mismatch is resolved and Vercel is building on the pinned pnpm 11.11.0 like everything else.

All four checks pass on `8653559` (same content as before, re-pushed to trigger a fresh deployment). Ready to merge whenever you are.

Worth noting the side effect now that the builder sees the whole repo: `pnpm install` covers all three workspace importers, so backend's and parsing's dependencies install during the web build too. Both lockfiles are consistent — CI installs them every run — so it's correctness-neutral, just a slower deployment.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #474, opened 2026-07-28
fix(web): actually tear down the test game-data server

Maelstromeous:
Fixes the "flaky" web suite that failed `Build & Test Web` four times today on #466 and #471 with every test passing.

It turned out to be **two independent bugs wearing one symptom**, and neither is a race in the test code.

## 1. The port-3001 server was never closed

```
close timed out after 10000ms
Tests closed successfully but something prevents Vite server from exiting
```

`hanging-process` named it in one word: **`TCPSERVERWRAP`** — the game-data server `globalSetup` starts, still listening after the run.

It was never closed because the teardown was wired up as `test.globalTeardown` in `vite.config.mts`. **That's a Jest option.** Vitest has no such thing — its config type exposes `globalSetup` only, it ignores the unknown key silently, and takes the teardown from what `globalSetup` *returns* (or a named `teardown` export):

```js
// vitest/dist/chunks/cli-api…js
else throw new Error(`invalid globalSetup file ${file}. Must export setup, teardown or have a default export`)
```

So `testing/global-teardown.ts` never ran. Not intermittently — never. `globalSetup` now returns the closer; it drops live connections first (`close()` waits on keep-alive sockets, and one would reintroduce the same hang) and skips teardown when port 3001 was already occupied, where the old code would have called `close()` on a server that never listened (`ERR_SERVER_NOT_RUNNING`). The dead file and the config key are deleted.

## 2. Work escaping the test that started it

That alone did **not** stop the random failures — two of three runs still exited 1, with all 1000 tests passing:

```
Vitest caught 5 unhandled errors during the test run.
EnvironmentTeardownError: [vitest-worker]: Closing rpc while "onUserConsoleLog" was pending
This error originated in "src/stores/app-store.spec.ts"
```

A `console.log` in flight over the worker RPC when the worker is torn down. Unhandled errors fail the run regardless of test results.

I instrumented `console.log` to dump a stack for anything logged after a file's tests finish. One run recorded **24 late logs**, every one rooted at the same place:

```
at app-store.ts:91                       ← tab-index watcher callback
at runAnimationFrameCallbacks (jsdom)
at Timeout._onTimeout         (jsdom)
```

The watcher defers the tab-switch load with `requestAnimationFrame` so the loading overlay paints first. **jsdom implements rAF on a ~16ms timer**, so under Vitest it defers the work past the end of the spec that switched tabs — the callback logs, then starts the `prepareLoader` → `setFactories` → `initFactories` cascade, which logs a dozen more times, all during worker teardown.

Two changes, both needed:

- `afterPaint` runs the callback inline in test mode, so the work happens inside the test that asked for it. Nothing paints under jsdom.
- `loadPause` makes the loader's 50ms/75ms-per-factory pauses 0 in test mode. Those pace the browser between factories so a large plan appears progressively; in tests they only keep the chain sleeping and logging after the spec ends. On its own this fixed nothing — the trigger was still deferred — which is why both are here.

## Verification

Three consecutive full runs, before and after. The bar is zero teardown errors, not three green exits — a green exit was always achievable by luck, and both "passing" diagnostic runs earlier printed the timeout while still exiting 0.

| | exit codes | teardown errors | tests |
| --- | --- | --- | --- |
| before | 1, 1, 0 | 4–6 per run | 1000 passed |
| after | 0, 0, 0 | **0** | 1000 passed |

Also gone: the hanging-handle dump, the 10s timeout, and "prevents Vite server from exiting". `Test server closed` now appears in the output for the first time. `lint-check` clean, and CI is green on all four checks here.

## Dead ends, for the record

Two hypotheses were tested and refuted by measurement before the instrumentation found the real one:

- **eventBus log volume** — `emit` is wrapped to `console.log` every event with its full payload (whole `Factory` objects) across the RPC. Gated under test mode: still exit 1, 4 teardown errors.
- **the loader pauses alone** — see above: necessary, but fixed nothing by themselves.


Maelstromeous:
Correction before this gets merged as a complete fix: **it isn't one.** CI is green here and the hang is genuinely gone, but running the suite three more times locally, two of the three still exited 1 — with zero timeout warnings and all 1000 tests passing.

The remaining cause is separate:

```
Vitest caught 6 unhandled errors during the test run.

⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
EnvironmentTeardownError: [vitest-worker]: Closing rpc while "onUserConsoleLog" was pending
This error originated in "src/stores/app-store.spec.ts"
```

A `console.log` is still in flight over the worker RPC when the worker is torn down, and unhandled errors fail the run regardless of test results. The passing run had none of these; the failing ones had six each.

So the picture is two independent bugs behind one symptom:

1. **The port-3001 server never closing** — fixed in this PR, and it was constant rather than intermittent.
2. **Late console logs racing worker teardown** — still open.

I'm testing a hypothesis for the second now: `src/utils/eventBus.ts` wraps `emit` to `console.log` every event *with its full payload* — whole `Factory` objects — so the suite pushes a large volume of serialized console traffic across the RPC, which is exactly the traffic that gets caught pending at teardown. If gating that log under test mode clears the unhandled errors across repeated runs, I'll add it here and update this PR; if it doesn't, I'll report what actually does.

Either way this PR's change stands on its own — the hanging handle was real and is fixed — but merging it will not stop the red jobs by itself.

---
_Generated by [Claude Code](https://claude.ai/code)_

Maelstromeous:
Following up on my correction above — the second cause is now found and fixed, and the PR description is rewritten to cover both.

It was **not** the eventBus log volume I guessed at. Instrumenting `console.log` to dump a stack for anything logged after a file's tests finish caught 24 late logs in a single run, every one rooted at the same callback:

```
at app-store.ts:91                       ← tab-index watcher
at runAnimationFrameCallbacks (jsdom)
at Timeout._onTimeout         (jsdom)
```

jsdom backs `requestAnimationFrame` with a ~16ms timer, so the tab-switch load gets deferred past the end of the spec that triggered it and then logs its way through the whole `prepareLoader` cascade during worker teardown.

Verification, three consecutive runs each way — the bar being zero teardown errors rather than three green exits, since a green exit was always possible by luck:

| | exit codes | teardown errors | tests |
| --- | --- | --- | --- |
| before | 1, 1, 0 | 4–6 per run | 1000 passed |
| after | 0, 0, 0 | **0** | 1000 passed |

CI is green on all four checks here too. This is now safe to merge as a complete fix — unlike when I first opened it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #486, opened 2026-07-28
fix(parser): derive mwPerItem from fuel energy, not a rounded burn rate

### LINKED ISSUE #485: Rocket Fuel demand for Fuel Generators is off by a rounding error (1000.001 / 2400.002 instead of 1000 / 2400)
Reported by Thomas [MRB] on Discord while building a 600 Oil → 2400 Rocket Fuel plant feeding 240 Fuel Generators at 240% clock.

The generators ask for `2400.002` Rocket Fuel/min instead of `2400`, so the user has to manually override the producing factory's Rocket Fuel amount to `2400` to get the two factories to line up. The existing near-integer snapping does not catch this.

## Reproduction

Share link (Thomas's plan): https://satisfactory-factories.app/share/putrid-dirty-secretary

Steps:
1. Create a new factory.
2. Add a Power Generator → Fuel-Powered Generator (Rocket Fuel).
3. Set the building count to `240` → fuel demand already shows **1000.001** (should be 1000).
4. Open the building group and set it to **240%** → fuel demand shows **2400.002** (should be 2400).

## Cause

Two separate things combine here.

**1. The parser derives `mwPerItem` from an already-rounded `perMin`.**

`parsing/src/recipes.ts:303-314` rounds the burn rate to 5 dp and then divides the building power by that rounded value:

```ts
primaryPerMin = parseFloat((burnRateMJ / primaryFuelPart.energyGeneratedInMJ).toFixed(5))
ingredients.push({
    part: fuelItem.primaryFuel,
    perMin: primaryPerMin,
    mwPerItem: building.power / primaryPerMin,
})
```

Rocket Fuel's true rate is 4.1666…/min, so it stores:

```json
{
  "id": "GeneratorFuel_RocketFuel",
  "ingredients": [
    { "part": "RocketFuel", "perMin": 4.16667, "mwPerItem": 59.999952000038405 }
  ],
  "building": { "name": "generatorfuel", "power": 250 }
}
```

`mwPerItem` should be exactly `60` (3600 MJ ÷ 60 s). The 8e-7 relative error is what leaks out:

- 240 gens @ 100% → 60,000 MW ÷ 59.999952 = 1000.0008 → **1000.001**
- 240 gens @ 240% → 144,000 MW ÷ 59.999952 = 2400.00192 → **2400.002**

`power.ts:135` does exactly this division (`producer.powerProduced / recipe.ingredients[0].mwPerItem`), so the error scales with plant size — 1200 gens @ 240% gives **12000.01**.

Only two power recipes in `gameData_v1.2-04.json` have a non-exact `mwPerItem`, both for the same reason:

| Recipe | stored `perMin` | stored `mwPerItem` | should be |
| --- | --- | --- | --- |
| `GeneratorFuel_RocketFuel` | 4.16667 | 59.999952000038405 | 60 |
| `GeneratorCoal_CompactedCoal` | 7.14286 | 10.499995800001681 | 10.5 |

Every other power recipe divides out exactly and is unaffected.

**2. Power producers never opt into `snapNearInteger`.**

`snapNearInteger` in `web/src/utils/numberFormatter.ts` exists precisely to absorb this class of drift, but `formatNumberFully`'s `snap` argument is opt-in and `power.ts` never passes it — `power.ts:118-135` calls `formatNumberFully(...)` with the default `snap = false` throughout, while `products.ts:96-99` does opt in. So the safety net that would have hidden `2400.002` is simply not wired up on the power producer path.

Note that snapping alone would not be a complete fix: the drift is proportional, so at ~5000+ Rocket Fuel/min it exceeds the 0.002 tolerance and would reappear.

## Suggested fix

1. **Parser (root cause):** derive `mwPerItem` from the unrounded rate — i.e. `primaryFuelPart.energyGeneratedInMJ / 60` — and keep the 5 dp rounding for the displayed `perMin` only. This makes Rocket Fuel `60` and Compacted Coal `10.5` exactly, and the division in `power.ts` round-trips cleanly. Needs a `dataVersion` bump in `web/src/config/config.ts` and a new `gameData_v<version>.json`.
2. **Frontend (defence in depth):** pass `snap = true` on the power producer quantities in `power.ts`, matching what `products.ts` already does, guarded the same way so user-set fractional clocks are still kept exact.

Maelstromeous:
Closes #485.

Fuel Generators on Rocket Fuel asked for **2400.002** fuel/min instead of 2400 on a 240-generator plant clocked to 240% (and 1000.001 instead of 1000 at 100%), forcing a manual override on the producing factory to make the two line up.

## Root cause

`parsing/src/recipes.ts` calculated `mwPerItem` as `building.power / primaryPerMin`, where `primaryPerMin` had already been rounded to 5dp to shed float noise. Rocket Fuel's true burn rate is 4.1666…/min, so `mwPerItem` stored `59.999952000038405` instead of exactly `60`.

`power.ts` divides a plant's total MW by that figure, so the 8e-7 relative error scaled with the plant:

- 240 gens @ 100% → 60,000 MW ÷ 59.999952 = **1000.001**
- 240 gens @ 240% → 144,000 MW ÷ 59.999952 = **2400.002**
- 1,200 gens @ 240% → **12000.01**

The 5dp rounding it depended on existed to hide noise the parser was creating itself: `burnRateMJ` was `(power / 60) / (1 / 3600)`, which turns 250 MW into 15000.000000000002.

## Fix

**Parser (root cause).** `mwPerItem` is now `energyGeneratedInMJ / 60`, exact by definition (3,600 MJ ÷ 60s = 60). The burn rate is derived with a single multiply (`power * 60`) instead of the divide-then-divide round trip, so the 5dp rounding is no longer needed and comes out — the stored `perMin` now round-trips against `mwPerItem`.

**Frontend (defence in depth).** Power producer quantities opt into the same whole-number snapping `products.ts` already uses, guarded the same way so a user-dialled fractional clock keeps its exact figures. The predicate moved out of `products.ts` into `factory-management/common.ts` as `hasFractionalClock`, shared by both paths. Snapping alone would not have been a complete fix — the old drift was proportional and exceeded the 0.002 tolerance past ~5,000/min — so the parser fix is what actually holds at scale.

## Game data scan

Swept every number in the generated data for float noise. Two categories turned up, both fixed:

| Value | Before | After |
| --- | --- | --- |
| `GeneratorFuel_RocketFuel` mwPerItem | 59.999952000038405 | 60 |
| `GeneratorCoal_CompactedCoal` mwPerItem | 10.499995800001681 | 10.5 |
| `GeneratorFuel_RocketFuel` perMin | 4.16667 | 4.166666666666667 |
| `GeneratorCoal_CompactedCoal` perMin | 7.14286 | 7.142857142857143 |
| `Alternate_SuperStateComputer` Computers | 7.199999999999999 | 7.2 |
| `Alternate_NuclearFuelRod_1` Crystal Oscillators / product | 0.6000000000000001 | 0.6 |
| `NuclearFuelRod` Encased Industrial Beams | 1.2000000000000002 | 1.2 |

The last three are a separate parser bug: production rates were `(60 / duration) * amount`, which loses precision on the divide. Multiplying first (`(60 * amount) / duration`) fixes all of them.

A full before/after diff of the regenerated data shows exactly these 8 values changed and nothing else. Everything still non-integer afterwards is a genuine repeating rate (1.6666666666666667 Crystal Oscillators/min) already at its closest representable value.

Data regenerated as `gameData_v1.2-05.json`; `dataVersion` bumped and the old file deleted.

## Tests

**Parser** (`parsing/tests/`):
- `mwPerItem` is exactly 60 / 10.5 for the two affected recipes.
- Every power recipe's `mwPerItem` equals its fuel's energy per second.
- Every fuel `perMin × mwPerItem` round-trips to the building's power.
- 240 generators at 100% and 240% divide out to whole 1000 / 2400.
- A sweep of **every number in the generated file** for float noise — a value within a rounding hair of a short decimal without being it. Genuine repeating rates pass; this is the guard against the whole class reappearing.
- The three production rates above assert their exact ratios.

**Frontend** (`web/src/utils/factory-management/power.spec.ts`): the issue's repro end to end — 240 Rocket Fuel generators demanding a whole 1000/min, then 2400/min once the group is clocked to 240%; 1,200 generators (the size where snapping alone could not save it); Compacted Coal; and a user-set fractional clock still producing an unsnapped 2400.01.

## Validation

- `parsing`: 38 tests pass (was 34).
- `web`: 1005 pass, 1 skipped, across 59 files.
- `pnpm lint` clean (no new warnings), `pnpm build` green including `vue-tsc`.
- Confirmed the new tests genuinely catch the bug: restoring the old `mwPerItem` values fails the 1,200-generator and fractional-clock assertions (5000.004 and 2400.012), with the smaller cases absorbed by the new snapping — i.e. each layer of the fix is covered by at least one test.


Maelstromeous:
Eyeballed correct on preview, requires recalc / value jiggle but works properly now.

PR #488, opened 2026-07-28
fix(web): repair rounding drift baked into saved plans on load

Maelstromeous:
Follow-up to #486, which fixed the source of the drift. This one goes back and repairs the plans that already have it baked in.

## Why #486 wasn't enough

Correcting the game data fixes everything the planner *recalculates*. But a plan's **seed** values — a product's amount, a generator's fuel or power figure, an import amount — are inputs to that calculation, not outputs. A plan saved holding 2400.002 keeps asking for 2400.002 forever, no matter how correct the recipe data becomes.

## What this does

`initFactories` now runs `repairPlanPrecision` over every loaded plan — all load paths (localStorage tabs, account sync, share links) funnel through it. It snaps seeds sitting a rounding hair off a whole number and forces the recalculation that clears the derived figures with them. It runs *after* the existing shape migrations so it can rely on products, producers and building groups being present.

Repaired seeds: `product.amount`, `powerProducer.fuelAmount` / `.powerAmount` / `.buildingAmount`, `powerProducer.ingredients[].perMin`, and `factory.inputs[].amount`.

## The tolerance scales with the value

This is the part that matters. #485's drift was *proportional* (8e-7 of the value), so it outgrew the runtime snap's flat 0.002 allowance past ~2,500/min — catching 12000.01 needs 0.012. `snapDriftedInteger` uses `max(0.002, |value| × 1e-6)`: identical to the existing `snapNearInteger` for small values, wide enough for large plants.

A value must still land within **one part per million** of the integer, which no hand-typed quantity does:

| Value | Repaired? |
| --- | --- |
| 1000.001, 2400.002, 5000.004, 12000.01 | yes |
| 12000.05 (4 ppm off) | no |
| 12000.5, 822.667 | no |
| 0.001 | no — never snaps to zero |

## Deliberate precision is respected

Same rule `calculateProducts` already applies: anything on a building group with a **user-set** fractional clock is skipped entirely, so 535.999 from a 223.333% clock stays 535.999. A solver-derived fractional clock doesn't count, matching the existing behaviour. An import inherits the precision of the factory supplying it rather than being judged on its own — the amount mirrors a quantity that factory decided on.

## Stale recipe figures

A power producer stores a copy of its recipe's ingredients, taken when the producer was created and **never refreshed** — so saves made before the fix still carry the old `59.999952` `mwPerItem`. Nothing reads it today (the calculation always goes to the live recipe), so this is currently inert; it's refreshed anyway rather than left as a known-wrong figure sitting in the save waiting for something to start reading it. Only Rocket Fuel and Compacted Coal producers are affected, so this doesn't force a recalculation on unrelated plans.

## Tests

The plan attached to the issue (shared as `putrid-dirty-secretary`) is captured verbatim as `485-drifted-plan.ts` and drives the tests — it's the 240-generator @ 240% repro, with 2400.002 in every Rocket Fuel field and the old `mwPerItem` on the producer.

- `repair.spec.ts` (15 tests): the issue's plan repairs to a wholly clean 2400 across the producer, its building group and the factory's parts ledger; the 240% clock survives; the pass is idempotent. Plus tolerance boundaries, the never-snap-to-zero guard, user-set vs solver-derived fractional clocks, and both input cases.
- `app-store.spec.ts`: the plan loads clean through `initFactories`, a repaired plan gets recalculated, and a plan needing no repair does **not** (the recalc is expensive on large plans, so this must not fire on every load).
- `numberFormatter.spec.ts`: `snapDriftedInteger` boundaries, including an explicit contrast showing `snapNearInteger` leaves 12000.01 alone where this doesn't.

## Validation

- web: 1026 pass, 1 skipped, across 60 files. parsing: 38 pass.
- `pnpm lint` clean (no new warnings), `pnpm build` green including `vue-tsc`.
- Confirmed the tests depend on the new code: stubbing out the `repairPlanPrecision` call fails the two app-store assertions (2400.002 instead of 2400, and no recalculation).

## Notes

- Repairs are logged to the console with a before/after line each, not surfaced in the UI. Happy to add a toast if you'd rather users were told their plan was touched.
- Export Calculator belt/pipe group amounts are user-set throughput figures and are deliberately left out of the sweep — they're not seeds for the calculation and the risk/benefit didn't justify it.


Maelstromeous:
Added the modal. Repairs are no longer console-only.

**What it looks like** — held back until loading finishes, so it doesn't land behind the loading overlay on the very load that produced it:

> **Plan data repaired**
>
> We have detected some data errors in your plan (usually to do with micro-rounding) and have fixed the following items:
>
> **FG TEST** / **Rocket Fuel** (Fuel rate)
> Fuel-Powered Generator (Rocket Fuel): 2400.002 → **2400**
>
> Nothing you need to do — your plan has already been recalculated with the corrected figures. If you had worked around this by hand, you can now undo that.

**Changes to get there:**

- Report entries are structured rather than pre-formatted strings, so the dialog renders friendly names (`Rocket Fuel`, `Fuel-Powered Generator (Rocket Fuel)`) instead of raw recipe ids.
- **Deduplicated.** A producer's `fuelAmount` and its fuel ingredient's `perMin` are the same number held twice — the first pass listed the same correction twice, which read as two separate problems. Identical corrections now collapse to one line.
- The report lives on the store rather than the event bus. A plan can be inited before the layout has mounted its listeners (the `factories` getter inits on first read), and a repair the user never sees reported is exactly what this is meant to prevent.
- `setup-vitest.ts` gained a `visualViewport` stub — jsdom has none and Vuetify's overlay reads the bare global while positioning itself, so any `v-dialog` component test dies on an unhandled `ReferenceError` without it. This unblocks dialog testing generally, not just here.

**Verification** — drove it in a real browser against the issue's plan seeded into localStorage: dialog appears with the right figures, dismisses cleanly, does not return, and no `2400.002` remains anywhere on the page afterwards. Plus 4 component tests covering the closed/open states and the rendered content.

web: 1032 pass, 1 skipped, 61 files. Lint clean, build green including `vue-tsc`.

Maelstromeous:
Dialog reworked, and there's now a button for it.

**Grouped and coloured** — one heading per factory (with the factory icon, sized above the item text), items bulleted beneath, old value in the semantic error red and new value in success green. Colours come from the `--sf-*` tokens rather than fresh literals.

```
🏭 Rocket Fuel Refinery
  • Rocket Fuel — Rocket Fuel (Quantity): 14400.012 → 14400
🏭 FG TEST
  • Rocket Fuel — Fuel-Powered Generator (Rocket Fuel) (Fuel rate): 2400.002 → 2400
  • Compacted Coal — Coal-Powered Generator (Compacted Coal) (Fuel rate): 3000.001 → 3000
  • Rocket Fuel — Imported from Rocket Fuel Refinery (Quantity): 2400.002 → 2400
🏭 Mega Rocket Fuel Plant
  • Rocket Fuel — Fuel-Powered Generator (Rocket Fuel) (Fuel rate): 12000.01 → 12000
  • Rocket Fuel — Imported from Rocket Fuel Refinery (Quantity): 12000.01 → 12000
```

**New template: "#485: Micro-rounding repair"** (debug-only, matching the other issue scenarios — so it shows on localhost automatically and stays out of production unless `?debug`). Three factories drifted three different ways: a product quantity, two generator fuel rates on different recipes, and two imports.

Every figure in it is what the old mwPerItem actually produced — `2880 × 250 / 59.999952 = 12000.0096`, `420 × 75 / 10.499996 = 3000.0012` — rather than numbers I invented. The Refinery (0.012) and Mega Plant (0.01) drifts sit **past the flat 0.002 snap tolerance**, so pressing the button exercises the scaling tolerance rather than just the easy end.

The verbatim capture from the issue stays untouched as `create485Scenario`; the demo plan builds on top of it, so the headline case is still real save data.

**Verified in a browser**: loading the template raises the dialog with three headings and six corrections, the computed colours resolve to `rgb(244, 67, 54)` / `rgb(76, 175, 80)`, no page errors, and after dismissing, the plan reads 14,400 / 12,000 with no drift text anywhere.

web: 1036 pass, 1 skipped, 61 files. Lint clean, build green.

Note the demo plan is deliberately unbalanced (the Refinery has no ingredient imports, the coal generator no coal supply) — it exists to show the repair, not to be a sensible plan, so the deficit chips it lands on are expected.

Maelstromeous:
image

PR #499, opened 2026-07-31
fix(web): stop factories advertising exports nobody imports

Maelstromeous:
Closes the "ghost exports" reports — a factory advertising an export to a factory that has no matching import, or exporting the wrong amount.

## How I approached it

I wrote an invariant checker first (requests ≡ inputs, in both directions, amounts included) and threw every mutation path at it before changing any code. The engine's happy paths turned out to be sound: deleting an import, deleting a product, retargeting an import, deleting a factory, and twelve seeded fuzz walks over the demo plan all reconciled correctly.

So the ghosts weren't coming from the calculation order. They were coming from data that got corrupt *outside* a calculation and then never got flushed. Every fix below has a failing test written before it.

## The five causes

**1. Factory IDs could collide.** `newFactory()` drew from `Math.floor(Math.random() * 10000)` with nothing checking whether the number was already taken — roughly a 1-in-6 chance of a collision at 60 factories, closer to 1-in-2 at 120. Export requests are keyed by the importing factory's ID, so a collision merges two factories as far as the dependency system is concerned: one request covering two different importers, the wrong factory named on the export chip, the supplier under-producing because it only sees one of the two demands, and deleting either factory taking the other's imports with it. It could also issue `0`, which `findFac` throws on and the input validation reads as "no factory selected".

IDs now come from `generateFactoryId(factories)`, issued against the plan and never 0. `addFactory` reassigns on collision so any future creation site is covered, and a plan loaded with a collision already in it has the clash broken and its exports rebuilt.

**2. An emptied request key was never pruned.** It renders as the importing factory's name under "Exports:" with no parts against it. `flushInvalidRequests` only deleted an emptied entry while it happened to be looking at it.

**3. Duplicate imports understated the export.** Only one request is ever raised per supplier + part, and `updateDependency` assigned the amount rather than adding to it — so importing 300 and 200 Iron Ingots from the same factory told the supplier only 200 were wanted. Amounts are now totalled, duplicates already in a saved plan are merged on load, and switching an import's factory onto a combination you already have clears the item so you pick again.

**4. A self-referencing input threw.** That aborted the calculation for every factory — a blank planner over one corrupt row. It's dropped with a warning now.

**5. A plan that was already wrong stayed wrong.** A plan whose figures look current is deliberately not recalculated on load (that's what makes tab switching fast), so a ghost saved by an older build survived every reload. `validateFactories` now returns whether the plan needs a recalculation, and `initFactories` honours it. Anyone already affected gets it cleared next time they open the planner.

## Also fixed

- **Deleting an import row took the other unfinished rows with it.** Rows were matched by factory + part, and every row you haven't finished filling in looks identical (no factory, no part) — so clicking the bin on one of three blank rows removed all three. Confirmed against `main` before and after (8 rows → 6 before, 8 → 7 after).
- The Imports `v-for` keyed on `${factoryId}-${outputPart}`, which is a duplicate key across unfinished rows; Vue can then patch the wrong row's selectors.
- A copied factory inherited the original's exports even though the importers are still buying from the original. The flush tore them down afterwards, but only after recalculating every factory involved.
- `getExportableFactories` pushed a factory once per exportable part, so a factory making three exportable items appeared three times.

## The contract

The rule that requests are derived from inputs is now written down once, in `web/src/utils/factory-management/dependency-integrity.ts`, and enforced from both ends: the app runs it when a plan loads (which is what triggers the repair recalculation), and the tests assert it after every mutation they make.

Alongside the per-case regression suite there's a seeded fuzz suite that walks the demo plan through 60 random edits — deleting imports and products, retargeting imports, adding, copying and deleting factories — re-checking the invariants after every one. A failure names the seed that reproduces it.

## Behaviour change worth calling out

`calculateAllDependencies` no longer throws on a self-referencing input, and the test asserting that it throws has been updated. Bricking the whole planner over one corrupt row is worse than dropping the row.

## Validation

- 1094 tests pass (63 files), `pnpm lint-check` and `vue-tsc --noEmit` clean.
- Drove the planner in Chromium through delete-import → add/delete blank rows → copy factory → delete factory → full reload, asserting the invariants against the persisted plan at each step. Clean throughout, and the blank-row fix verified against `main` for contrast.

## Not in this PR

Every factory edit throws an uncaught `DataCloneError: Failed to execute 'structuredClone'`. I confirmed it's pre-existing by stashing this work and re-running the same browser walkthrough on `main` — identical. Calculations still complete, so something is swallowing it. Being investigated separately.


Maelstromeous:
Pushed `e826b9f`, which resolves the `DataCloneError` I flagged as out of scope above — it turned out to be worse than a stray console error.

**Deleting a factory that supplies another one silently stopped half way.** The plan was never recalculated afterwards and `regenerateSortOrders` never ran: deleting the first of nine factories left the rest numbered 1–9 instead of 0–8.

`cloneForCalculation` copies the plan with `structuredClone`, which refuses to copy a Proxy. Taking `factory.inputs.filter(...)` off a *reactive* factory reads the rows out through the proxy, so assigning the filtered array back stores those proxies as the new elements — and `removeFactoryDependants` does exactly that to every dependant of the factory being deleted. Everything after the clone in the click handler was then skipped by the throw.

It only surfaces when a dependant still has another import afterwards; an emptied array has no elements left to be proxies, which is why it hid so well.

Fix is in two parts:
- Every site that rewrites an array derived from a reactive one goes through `rawArray()` (new, in `factory-management/common.ts`), which unwraps the elements first.
- `cloneForCalculation` falls back to a deep unwrap, with a `console.error` naming the cause, rather than abandoning the calculation if a future site misses it.

`reactivity.spec.ts` runs against reactive factories the way the app does, reproduces the original failure, and asserts the fallback never actually has to fire — so the real fix stays load-bearing rather than the net covering for it.

Verified before/after in the browser: on `main` the same walkthrough throws and leaves `displayOrder` at `[1..9]`; on this branch there are no page errors and it's `[0..8]`. 1102 tests pass, lint and `vue-tsc` clean.

Maelstromeous:
Split the `DataCloneError` fix out into #500 — it's independent of the ghost-export work and worth reviewing on its own. This PR is now based on that branch, so its diff shows only the ghost-export changes; it'll retarget to `main` automatically once #500 merges.

The dependency runs that way round: this PR adds two more reactive-array assignments of its own, so it needs #500's `rawArray()` to exist.

Worth noting from the split — carving the fix back onto `main` made `reactivity.spec.ts` fail on a site that only exists there, `deleteInputPair`. So deleting an *import* (not just a factory) poisons the plan on `main` too, which is the more common action of the two. That's covered in #500.

Maelstromeous:
Pushed `fbd46fc` — the repair reporting is now generic, and load-time detection covers the whole chain.

## One list, one dialog

`PlanRepair` is a union: the existing quantity entry (drifted micro-rounding) plus a new `StructuralRepair` carrying a plain-language summary. `validateFactories` returns what it repaired instead of a boolean; `initFactories` merges that with the precision repairs into `planRepairs`; `PlanRepairDialog` renders both kinds grouped by factory.

**The browser alert is gone.** It fired before the planner had drawn anything, said nothing useful, and sent you to the dev console. The console line stays — it carries the factory IDs, which are noise to the user but the first thing anyone debugging a shared plan asks for.

## Load-time detection of the whole chain

`validateFactories` gained `repairDependencyChain`, so a loaded plan is checked end to end rather than relying on the recalculation to flush it:

- exports nobody imports
- export entries emptied of their items (these render as a factory name with nothing under it)
- export amounts that disagree with what the importer asks for
- imports the supplier has no record of

Plus what was already there: colliding factory IDs, duplicate import rows, self-imports, imports from deleted factories, zero/negative quantities.

## The invalid-plan fixture

`499-broken-chain-plan.ts` is built and calculated **correctly first**, then damaged — so its derived figures still look current and nothing would otherwise recalculate it, which is exactly how a real corrupt save behaves. It carries a copied factory holding the original's exports, two factories sharing an ID, a ghost export, a drifted export amount, duplicate import rows, an import the supplier doesn't know about, an import from a deleted factory, a self-import, and an emptied export entry.

`plan-repairs.spec.ts` loads it through `appStore.initFactories` and asserts, for each fault, that it is **both corrected and reported** — and that loading the repaired plan a second time finds nothing left to do.

## Validation

1126 tests pass (65 files), lint and `vue-tsc` clean.

Verified in Chromium by damaging the demo plan in `localStorage` (ghost export, ID collision, drifted export amount) and reloading. The dialog opens with no browser alert and no page errors:

> **Copper Basics** — Shared an internal ID with another factory, which mixes up their imports and exports. It has been given an ID of its own; check its imports still point where you expect.
> **Oil Processing** — Was set to export 397/min of Plastic to "Circuit Boards" while that factory asks for 320/min. The export has been corrected.
> **Iron Rods** — Imports CopperIngot from "Copper Ingots", which had no record of supplying it. The export has been restored.

The ID collision cascades into several further repairs, which is the honest outcome — once two factories share an ID the plan genuinely cannot tell which one supplied what, and each step of the unpicking is reported rather than done quietly.

Maelstromeous:
Pushed `0b0c910` — the **#485** template now exercises both kinds of repair in one load.

It's the same template, expanded rather than a second one: **`#485/#499: Plan repair`** in the Templates dialog (debug mode). Alongside the drifted quantities it now carries a broken chain — a copied Refinery inheriting the original's exports, an export that no longer matches its import (3,200 against an import of 2,400), an import the supplier has no record of, an export entry pointing at nothing, and two deliberately unlinked factories sharing an internal ID.

Loading it raises one dialog with both kinds listed together, and the Refinery and its copy each show a quantity repair *and* structural repairs under a single heading:

> **Rocket Fuel Refinery**
> • Was set to export 3200/min of Rocket Fuel to "FG TEST" while that factory asks for 2400.002/min. The export has been corrected.
> • Was exporting to a factory the plan can no longer identify. The export has been removed.
> • **Rocket Fuel** — Rocket Fuel (Quantity): 14400.012 → 14400

The ID collision is on two factories with no links, deliberately — inside the chain it cascades (the plan genuinely can't tell which factory supplied what) and would bury everything else. Noted in the template description.

Repair summaries now use part display names ("Rocket Fuel", not "RocketFuel") to match the quantity lines they sit beside.

## A real bug fell out of building it

**A plan that has never been calculated had its valid imports and exports torn up on its first calculation.** No calculation means no part ledger, and two checks read an empty ledger as "this factory doesn't make that":

- `flushInvalidRequests`' part and product checks
- `calculateFactoryDependencies`' provider-part check

`loadMode` guarded the second one, but `deleteRequestPair` recalculates *mid-flush* without passing it — so the guard leaked and the input was deleted anyway. Both now hold off until there is a ledger to read. It's latent on `main` (saved plans always carry parts, and no template shipped exports), but any template or code-built plan with exports would hit it — which is exactly what this one is. Two regression tests added under "a plan that has never been calculated".

1128 tests pass, lint and `vue-tsc` clean. Verified end to end in Chromium: template loads, dialog lists everything above, and afterwards the plan has unique IDs, no chain violations, and whole-number quantities (Refinery 14,400 / FG TEST import 2,400 / Mega Plant 12,000).

Issue #169, opened 2024-11-25
Add Planner Factory Folders


Maelstromeous:
For some people's plans (which is a LOT of factories) we've recieved complaints that managing a vast number of factories via a list system is a bit combersome when having to scroll around.

So, as a QoL feature, we should add Plan Folders which allows the user to place a factory within a folder and that folder has the option to be collapsed, just like a factory has now.

This Folder should also display the production metrics of a Factory Group in order to have a proper use. This would also resolve some other issues requesting metrics e.g. #101. We could place all current factories in a "default" group, which would generate some statistics, and allow the user to migrate a factory from the default group to one of their choosing, which shows its own metrics.

This data structure wise is simply adding `factory.group = 1` to the data structure.

Hard part will be the UI.

Maelstromeous:
This is basically just tabs but with extra steps.

Maelstromeous:
Reopening this. Mines have just landed — raw resources can now be extracted as products, either on site or as dedicated mine factories — and that pushes plans towards having a lot more factories, many of which are one-line mines I don't want to scroll past. Being able to classify a factory (e.g. as a mine) and collapse that group in the sidebar is the missing half of it.

Per-factory icons would help for the same reason, but that's a separate ask and shouldn't block this.

Maelstromeous:
Done in https://github.com/satisfactory-factories/application/pull/514

PR #503, opened 2026-07-31
feat: mining, resource wells, and an end to assuming raw inputs

Maelstromeous:
## Manual steps on merge

**None.** No environment variables, secrets, migrations or scripts. `dataVersion` moves to `1.2-07`, which is a code change and re-downloads on its own — nothing to run, and the deploy is safe without intervention.

---

Raw resources were an unspoken assumption: `calculatePartRaw()` silently topped up any shortfall, so a plan could never tell me it was short of ore, and there was no way to say "this ore comes from twelve Mk.3s on pure nodes". This makes extraction a first-class product and removes the assumption altogether, with a wizard to migrate the plans that relied on it.

## Extraction

Extractors sit on resource nodes and never appear in `mProducedIn`, so nothing in the game data links them to what they produce. The parser synthesises a recipe per resource: Miner Mk.1/2/3 for the ten solid node resources, the Oil Extractor for Crude Oil, the Water Extractor for Water, and Resource Well Pressurizers for Water, Crude Oil and Nitrogen Gas. Nitrogen Gas becomes extractable for the first time.

**Miner mark and node purity are set per building group**, because one ore line routinely mixes a Mk.3 on a pure node with a Mk.2 on a normal one. The trick that avoided re-deriving the solver: mark × purity becomes an *output multiplier relative to the recipe's reference rate* (Mk.1 at normal), which is the same shape somersloops already use. A Mk.3 group on pure nodes simply counts as 8 effective buildings, so `syncBuildingGroups`, `bestEffortUpdateBuildingCount` and the rest work unchanged.

Overclock composes on top, and the numbers match the game: a Mk.2 on a pure node gives 240/min @ 15 MW at 100%, 560/min @ 46 MW at 233.3333%, and 600/min @ 50.4 MW at 250%. Purity changes the yield, never the power.

## Resource wells

A well breaks two assumptions — the powered building isn't the one extracting, and satellites within one well have mixed purities. But the pressurizer's clock scales every satellite proportionally, which is exactly what a group's clock already means. So one group is one well: `buildingCount` is the pressurizers, and the group carries how many satellites stand on impure, normal and pure micro-nodes. Verified against a real well: 1 normal + 6 pure reads 780 m³/min at 150 MW, and 1,950 m³/min at 503.7 MW at 250%. Satellites are counted as buildings to place but draw nothing.

## Raw inputs — BREAKING

**Raw resources are no longer assumed to be supplied.** They are mined or imported like anything
else, and there is no setting. Plans built before this show shortages where supply was previously
assumed — nothing is changed or lost, the plan is only being honest about what was already missing.

The switch I added earlier in this branch is gone, along with the per-factory override, the plan
default and `factory-management/settings.ts`. An optional assumption meant the same plan could mean
different things to different people with nothing on screen saying so, and every feature downstream
of raw supply had to ask which mode it was in. The bug that ended it: **a factory mining 100 of the
180 limestone it needed read as fully satisfied**, because `rawChipReason` returned `extracted`
before it could ever report the assumed 80. Extraction, importing and gathering are independent
facts about a part, so they are independent predicates now rather than one mutually exclusive enum.

**The assumption did not leave the engine, it became data-driven.** `calculatePartRaw` tops a part
up if and only if nothing in the game data can extract it:

```ts
partData.amountSuppliedViaRaw = getHandGatheredParts(gameData).has(part) ? shortfall : 0
```

That is 11 of the 24 raw resources — Leaves, Wood, Mycelia, the four alien remains, the three power
slugs and the FICSMAS Gift, none of which has a producing recipe of any kind. Demanding those be
mined would leave every Fabric, Charcoal, Biomass and Protein factory permanently red with nothing
the planner could offer, so they stay satisfied behind an amber **Manually gathered** chip. The
factory stays green: there is nothing to build and nothing to fix.

**The rule counts resource wells, deliberately.** `getExtractionRecipeForPart` excludes them, and
reusing it here would class well-only Nitrogen Gas as hand-gathered and silently erase every
Nitrogen shortage in every plan. Two questions that look identical and are not — "does an extractor
exist at all" versus "can one be created automatically". `parts.spec.ts` pins the exact 11 ids,
because game data is regenerated and a new recipe could otherwise reclassify one unnoticed.

Downstream nothing else needed a guard: hand-gathered parts leave the engine satisfied, so
`status.ts`'s `rawShortage` and every shortage-button predicate are already excluded by their
existing `!part.satisfied` check.

## The Raw Resources Wizard

Removing the assumption without help would be cruel, so the same change ships the migration tool. A
row per factory and unmet raw part, each choosing: a shared mine factory, extraction on the spot, an
import from a factory that already mines it, or nothing. It lives in **Options** permanently — it is
as useful the tenth time a new factory comes up short as it is on migration day.

- **One mine per resource for the whole plan**, sized to everything that asked for it. A plan short
  of iron in eight places gets one Iron Ore Mine, not eight.
- **Water defaults to on-site**; a resource already mined somewhere defaults to importing from there
  rather than building a second mine next to the first.
- **Wells are the one thing it will not build.** A well's rate comes from its satellites, so solving
  a target rate against a fresh one multiplies the *pressurizer* and lands an order of magnitude out
  while reading as a solved plan. Nitrogen rows offer an import or nothing, and say why.
- **Apply is transactional.** `appStore.addFactory()` calls `schedulePersist()` on every invocation,
  so incremental mutation would leave orphan mine factories in the saved plan if anything threw —
  and there is no undo. `applyRawWizard` clones, validates every row against the live plan, mutates
  the clone, calculates it, and returns it for `setFactories()` to commit in one pass. The cost is
  that the wizard owns the ID-collision repair that normally lives in `addFactory()`; factory IDs
  key every export request, so a collision would cross two factories' wires silently.
- **The summary is the run that gets committed**, not a second count of the same rows, so it cannot
  promise something the apply does not do.
- Rows carry the shortfall captured when the table was read and revalidate it at apply time, so a
  stale row aborts rather than quietly re-pricing.

### The confirmation screen

Counts alone were not enough to consent to: "20 imports" says nothing about which factories are
about to change. It now lists every factory the run touches, and per item what that factory will
produce, what it imports and where each of it goes — read off the *calculated* clone rather than
predicted from the rows, because one mine feeding six factories is one product and six exports, and
only the finished dependency pass knows that. Imports matter most here: a factory the wizard only
wires up has no change to its products at all, so without them it was labelled "modified" with
nothing to point at.

Every line says what the run did to it — created, or raised what was already there — recorded as
the change is made, since afterwards a raised amount is indistinguishable from one that was always
that size. A factory reads as **New factory** or **Modified** and carries its colour through the
whole block.

- **Where new mines land is a choice** — top of the plan or bottom, defaulting to top. Eleven mines
  appended to the end of a long plan are eleven factories you have to go looking for. Changing the
  answer reorders the already-built result rather than re-running the wizard, which would rebuild
  the mines and discard anything renamed.
- **Mines can be renamed before they are built**, from a pencil beside the name. The wizard's own
  names are the ones most worth changing, and this is the last moment before they are committed;
  the rename follows through to every import and export reference on the screen.
- **A backup is one button.** Telling someone to copy their plan to the clipboard is telling them to
  solve the problem themselves, at the one moment they are about to lose it. The button writes the
  plan exactly as it stands, in the shape **Paste plan** already reads — through a shared serializer
  so the download and the existing copy-to-clipboard cannot drift, since a backup the paste path
  cannot read is only found out at the point of needing it.

`addShortageToFactory` now takes a **required, validated** `amount` rather than re-reading
`amountRemaining`, and loses its `Math.abs` — a negative there meant a surplus, and turning that
into production is exactly the bug an `abs()` would hide. Both existing callers pass what the
function computed for them before, so their behaviour is unchanged.

**Mines are built on Mk.2s at normal purity.** `addProductToFactory` creates the building group on
the recipe's *reference* extractor, so swapping in a Mk.2 afterwards doubles the group's output
against an unchanged amount — a brand new mine already in building-group mismatch. The mark is
applied with group sync temporarily on so the amount drives the count, then restored to the unsynced
default mines are created with. Measured, not assumed; the table is in `.claude/memory/`.

## Smaller things in here

- The Converter's ore recipes are relabelled "Iron Ore (Convert: Limestone)" — they otherwise sat beside "Iron Ore (Miner)" reading as though the ore just came from limestone. All seventeen of them.
- Each raw resource in Imports gets a "mine it here" button that adds the extraction at exactly the shortfall.
- The Demo plan gains a **Raw Materials Mine** hosting Limestone, Sulfur and Uranium for the nuclear chain, and Uranium Power pumps its own water from 22 Water Extractors — so the demo now covers all three shapes extraction takes (dedicated single-resource mine, multi-resource mine, and on-site).
- A mine's building groups report **Effective Output** rather than effective buildings, which were counted in Mk.1-equivalents and meaningless when you're building Mk.3s. When short, a second row lists what would cover it at every purity, counted in buildings at 100% — so a gap a quarter of an extractor wide reads `0.25` rather than rounding up to a whole one that would overshoot.
- Mines start with building group sync disabled: the first thing anyone does to a new mine is swap the default Mk.1, and with sync on that overwrites the quantity they just typed.
- Satisfaction gains an **Exported** chip beside Product and Imported, so what leaves a factory reads off the item's own row rather than only off the Exports column.
- The satisfaction chips lead with the icon their concept wears elsewhere in the app. Some of those names didn't exist in the Font Awesome build this app vendors, which is FA5-era: an unknown name still renders an `<svg>`, just filled with a placeholder, so nothing errors and nothing shows. Fixing that also turned up a World Data close button that had been invisible long enough to carry a comment wondering why.
- Node purity and satellite counts get their own semantic colour — they were using the raw-resource beige, i.e. the colour of the ore itself.

## Validation

1,346 web tests and 50 parser tests pass; lint and build clean. The rate and power figures are asserted against the wiki rather than derived from our own code, so a regression shows up as a number the game would disagree with. Also covered: a mine exporting ore to a smelter, mining and smelting on site, a well feeding a consumer, a save/load round trip preserving mark/purity/satellites, and that a no-op recalculation still performs **zero reactive writes** with extraction in the plan.

The wizard's specs lead with the destructive cases rather than the counts, because those are what decide whether saved data survives: a failing row writes **nothing**, the input plan is byte-identical after an apply, new mine IDs cannot collide, a well-only row cannot be forced to mine, a moved shortfall aborts, and an existing *non-extraction* product for the same raw is never bumped.

Driven end to end in a browser as well. On the Demo plan: the breaking-change notice appears, the wizard collects 4 rows (Water preselected on-site, the rest to mines), the summary reads "3 mine factories, 4 extraction products, 3 imports", and after applying, Limestone/Sulfur/Uranium Mines exist at 180/160/200 per minute on `2x minermk2@normal` with **no building-group mismatch** and sync back off. No raw shortage remains anywhere in the plan.

On the MegaPlan, which is the worst case: 31 changes across 19 factories, 11 mines created, unique
IDs throughout, **zero building-group mismatches**, and only the three Nitrogen rows left short.
Applying with mines at the top and again at the bottom both leave the plan intact and the mines
where they were asked for. The backup was round-tripped for real rather than assumed — downloaded
from the review screen (185 KB, 31 factories), then fed back through **Paste plan**, which restored
the pre-wizard plan exactly.

Also verified directly: `fa-hands` resolves to a real 698-character path rather than the 1408-character placeholder this Font Awesome build renders for unknown names (checked against a deliberate miss as a control), and `.sf-chip.hand-gathered` resolves to amber `#f57f17`.

Several of the wizard's UI defects were found by measuring rather than looking, and the measurements
run both ways so they can fail: the radio hit area was off by Chrome's UA `input[type=radio]` margin
(clicking the centre of a row did nothing while the bottom worked, confirmed with real mouse events
rather than synthetic `.click()`, which would have passed against the broken version); the review
table's four columns each set their own line height, so 141 of 141 rows were misaligned before the
fix and none after; and the group accent stripe was painting through the transparent border that
makes the gap between factories, leaving stubs of colour floating in the space between them.

## Notes for review

- `dataVersion` moves to `1.2-07`. It caches in localStorage, so it has to change whenever the data is regenerated or clients keep serving the old JSON.
- **Known and deliberate:** a product's declared `amount`, not its building groups' actual output, is what feeds downstream factories. An under-built mine therefore still satisfies its consumers, though its own card goes red with a `buildingGroupMismatch`. This is how building groups have always worked — any product with two or more groups already behaves this way on `main` — but mines defaulting to sync-off makes it reachable without choosing it. Left alone here because the fix is engine-wide.
- **The Simple and MegaPlan templates now load with raw shortages**, because they draw resources nothing in them digs up. Left as-is: it is the honest state and gives the wizard something to demonstrate. The Demo template *was* fixed, since it is the first thing a new user sees — it now mines everything, and what it is still short of is only the bottleneck and missing parts it was always meant to show. The Mining template is unaffected.
- `status-regression.spec.ts` had assumed the MegaPlan template was clean, which is how it guarded against false positives. It isn't any more (it draws ore it doesn't mine), so that flag flipped; the real assertion — that the old and new rollups flag the same factories name for name — still passes untouched.
- Wells inherit the sync-off default, so describing a well's satellites doesn't move the product quantity — it flags as over-producing instead. Consistent with the mine reasoning, but a well's satellites are a description of the world rather than a choice, so I may want wells specifically to keep sync on.
- The backend's copies of `Recipes.ts` / `FactoryInterface.ts` are marked as mirroring web but have drifted well behind already; I've left them alone since the backend stores plans as blobs.

Closes part of the mining request; factory folders (#169) and per-factory icons are deliberately out of scope.


Maelstromeous:
The commits from this branch are now in `update6`, which goes to `main` via #512 — that PR carries the full description and is where this work is reviewed from here.

I tried to retarget this PR's base to `update6` so it would resolve there, but GitHub refuses: there are no new commits between the two branches, since the merge already happened. Leaving it open rather than closing it, so it marks itself **merged** when `update6` reaches `main` instead of being recorded as abandoned.

Maelstromeous:
Merged into #512

